PR #10, opened 2017-06-28
utils: Fix wrong copy arguments

sboeuf:
The flag --preserve=all was not properly enabled through the code.
This patch fixes this issue.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -21,7 +21,7 @@ import (
 	"os/exec"
 )
 
-const cpBinaryName = "cp --preserve=all"
+const cpBinaryName = "cp"
 
 func fileCopy(srcPath, dstPath string) error {
 	if srcPath == "" {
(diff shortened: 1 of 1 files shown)


mcastelino (commented):
LGTM. 

Longer term we should evaluate elimination of the exec and replace it with a combination of ioutil.WriteFile/ReadFile with permission preservation. 

It will remove the dependency on the contents of the roots.

sboeuf:
@mcastelino I agree but we had the same discussion with @sameo and @jodh-intel for virtcontainers or the runtime, and we agreed it is safer to use cp because this is a very complex command. I think @jodh-intel did point to me a comment from one Go's maintainer saying that.
But we can still discuss again and try to implement our own file copy, I agree it would avoid any dependency on the guest rootfs/image.

PR #14, opened 2017-06-28
network: Rely on netlink handler to improve performance

sboeuf:
Instead of opening/closing the netlink socket everytime we need
to perform a netlink action, we rely on a netlink handler which
keeps the socket opened until we decide to close it.


mcastelino (commented):
LGTM.

PR #41, opened 2017-06-30
network: Support more IP and netmast formats

### LINKED ISSUE #40: network: Support different IP destination and mask formats
In order to make this agent compatible with both expectations from Clear Containers 2.1 and 3.0, and based on hyperstart agent, we have to accept several formats for the network mask and the destination IP address for a route.

sboeuf:
In order to get Clear Containers 2.1 working, we have to accept
slightly different formats from what we were previously expecting
regarding the netmask and the destination IP of a route.

About the routes, we now accept the destination IP without the
netmask attached, meaning the agent has to find it and concatenate
it to the IP provided.

About the interfaces, if the netmask is in IP format 255.255.0.0
for instance, the agent will ignore it and will rely on the IP
address to determine the netmask in decimal format.

Fixes #40

### DIFF
--- a/network.go
+++ b/network.go
@@ -44,6 +44,17 @@ func findLinkFromHwAddr(netHandle *netlink.Handle, hwAddr string) (netlink.Link,
 	return nil, fmt.Errorf("Could not find the link corresponding to HwAddr %q", hwAddr)
 }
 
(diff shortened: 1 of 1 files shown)


sboeuf:
Tested with 2.1 and 3.0. Both are working fine.

PR #59, opened 2017-07-22
channels: Stat /sys/class/virtio-ports/*/name before reading it

### LINKED ISSUE #58: Don't try to read /sys/class/virtio-ports/*/name if it does not exist
When trying to open the CTL and TTY channels, we should not try to open any file /sys/class/virtio-ports/*/name with this path without checking the path before. Indeed, we concatenate `name` to the prefix `/sys/class/virtio-ports/*`, but we don't want to fail in case the file does not exist because the ports we need can be any `vport*p*` from `/sys/class/virtio-ports/`.

sboeuf:
We don't want to end up getting an error because we tried to read a
non existing file. Indeed, we want to scan every entry inside the
path /sys/class/virtio-ports/, without throwing an error because
on of those entries does not have a file named "name". In that
case, we just want to skip it.

Fixes #58

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -396,6 +396,15 @@ func findVirtualSerialPath(serialName string) (string, error) {
 
 	for _, port := range ports {
 		path := filepath.Join(virtIOPath, port, "name")
+		if _, err := os.Stat(path); err != nil {
+			if os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


mcastelino (approved):
LGTM

PR #63, opened 2017-07-27
agent: Kill only init process if specified this way

### LINKED ISSUE #61: Fix Semaphore CI
We have to fix our scripts so that we can properly run our CI on each PR.

sboeuf:
In case we were running CRI-O testing inside Semaphore, because the
system is kind of slow, the redis container process takes some time
to really start. Because we were issuing a kill signal right after
the start returned (does not mean the container's process finished
to start), and at the same time we were trying to kill all processes
related to the init process, this was hanging. Indeed, we should not
try to kill all processes until the init one is properly running.

This patch relies on AllProcesses flag provided in the "killcontainer"
payload, instead of hardcoding this value to "true". Indeed, the CRI-O
tests explicitely ask for a kill without the --all falg enabled. This
simple change allows our agent to get CRI-O testing passing properly
on Semaphore CI.

Fixes #61

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -1036,7 +1036,8 @@ func killContainerCb(pod *pod, data []byte) error {
 		return fmt.Errorf("Container %s not found, impossible to signal", payload.ID)
 	}
 
-	if err := pod.containers[payload.ID].container.Signal(payload.Signal, true); err != nil {
+	// Use AllProcesses to make sure we carry forward the flag passed by the runtime.
+	if err := pod.containers[payload.ID].container.Signal(payload.Signal, payload.AllProcesses); err != nil {
 		return err
 	}
 


jcvenegas:
lgtm

sboeuf:
@amshinde, @mcastelino please take a look.

jodh-intel:
The change looks fine, but please can you add a brief comment in the code explaining why `AllProcesses` is being used.

Also, there are a couple of typos in the commit (falg => flag, explicitely => explicitly).

grahamwhaley:
The patch looks fine to me, but I do wonder if we are still glossing over a race/bug condition - presumably somebody can still issue a start immediately followed by a kill-all and it could hang up. So, is there a deeper rooted issue here still that should still be fixed some time in the future?

sboeuf:
@grahamwhaley yes I agree, and that's why I am planning to start a discussion with CRI-O/runc folks.

sboeuf:
@jodh-intel I am fixing the typo, but I don't think we need any comment to explain why we use `AllProcesses`. I mean, it is obvious that we are passing the flag that we got from our runtime, and that's the same flag expected by libcontainer. What comment would you expect ?

jodh-intel:
@sboeuf - question: why wasn't it obvious to use it in the previous version of the code then? :)

mcastelino:
LGTM

[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/63/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent) [Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/63/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

mcastelino (approved):
LGTM

PR #70, opened 2017-08-04
Add  timestamp to agent commands

jcvenegas:
- Agent now can get log level configuration from cmdline.
- Changed logrus time format to : `TimestampFormat: time.RFC3339Nano` to get nanoseconds precision.
- Start and end of commands is logged with the message:   `{event}_start` or  `{event}_end`
- An special case is logged `agent_started seconds.nanoseconds` this get uptime with good clock precision.  
- Logrus format is JSON for easy parsing

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -161,7 +161,7 @@ func init() {
 		runtime.LockOSThread()
 		factory, _ := libcontainer.New("")
 		if err := factory.StartInitialization(); err != nil {
-			agentLog.Infof("init went wrong: %v", err)
+			agentLog.Errorf("init went wrong: %v", err)
 		}
 		panic("--this line should have never been executed, congratulations--")
 	}
@@ -230,7 +230,7 @@ func (p *pod) controlLoop(wg *sync.WaitGroup) {
 				continue
 			}
 
-			agentLog.Infof("Read on ctl channel failed: %v", err)
+			agentLog.Errorf("Read on ctl channel failed: %v", err)
 			break
 		}
 
@@ -707,7 +707,7 @@ func (p *pod) runContainerProcess(cid, pid string, terminal bool, started chan e
 
 	processState, err := p.containers[cid].processes[pid].process.Wait()
 	if err != nil {
-		agentLog.Infof("Wait for process %s failed: %v", pid, err)
+		agentLog.Errorf("Wait for process %s failed: %v", pid, err)
 	}
 
 	// Close pipes to terminate routeOutput() go routines.
--- a/agent.go
+++ b/agent.go
@@ -63,6 +63,7 @@ const (
 	optionPrefix       = "agent."
 	logLevelFlag       = optionPrefix + "log"
 	defaultLogLevel    = logrus.InfoLevel
+	cannotGetTimeMsg   = "Failed to get time for event %s:%v"
 )
 
 var capsList = []string{
@@ -180,6 +181,13 @@ func main() {
 	applyConfig(config)
 
 	agentLog.Infof("Agent version: %s", Version)
+
+	if uptime, err := newEventTime(agentStartedEvent); err != nil {
+		agentLog.Errorf("Failed to get uptime %v", err)
+	} else {
+		agentLog.Infof("%s", uptime)
+	}
+
 	// Initialiaze wait group waiting for loops to be terminated
 	var wgLoops sync.WaitGroup
 	wgLoops.Add(1)
@@ -808,7 +816,12 @@ func (p *pod) runCmd(cmd hyper.HyperCmd, data []byte) error {
 		return fmt.Errorf("No callback found for command %q", hyper.CmdToString(cmd))
 	}
 
-	return cb(p, data)
+	agentLog.Infof("%s", hyper.CmdToString(cmd)+"_start")
+
+	cbErr := cb(p, data)
+
+	agentLog.Infof("%s", hyper.CmdToString(cmd)+"_end")
+	return cbErr
 }
 
 func startPodCb(pod *pod, data []byte) error {
--- a/agent.go
+++ b/agent.go
@@ -59,6 +59,10 @@ const (
 	statelessPassword  = "/usr/share/defaults/etc/passwd"
 	defaultGroup       = "/etc/group"
 	statelessGroup     = "/usr/share/defaults/etc/group"
+	kernelCmdlineFile  = "/proc/cmdline"
+	optionPrefix       = "agent."
+	logLevelFlag       = optionPrefix + "log"
+	defaultLogLevel    = logrus.InfoLevel
 )
 
 var capsList = []string{
@@ -100,6 +104,10 @@ var capsList = []string{
 	"CAP_SYS_TTY_CONFIG",
 	"CAP_SYSLOG",
 	"CAP_WAKE_ALARM",
+}
+
+type agentConfig struct {
+	logLevel logrus.Level
 }
 
 type process struct {
@@ -163,6 +171,14 @@ func init() {
 var Version = "unknown"
 
 func main() {
+
+	agentLog.Formatter = &logrus.JSONFormatter{TimestampFormat: time.RFC3339Nano}
+	config := newConfig(defaultLogLevel)
+	if err := config.getConfig(kernelCmdlineFile); err != nil {
+		agentLog.Warnf("Failed to get config from ernel cmdline: %v", err)
+	}
+	applyConfig(config)
+
 	agentLog.Infof("Agent version: %s", Version)
 	// Initialiaze wait group waiting for loops to be terminated
 	var wgLoops sync.WaitGroup
@@ -1209,4 +1225,67 @@ func winsizeCb(pod *pod, data []byte) error {
 	}
 
 	return nil
+}
+
+func newConfig(level logrus.Level) agentConfig {
+	config := agentConfig{}
+	config.logLevel = level
+	return config
+}
+
+//Get the agent configuration from kernel cmdline
+func (c *agentConfig) getConfig(cmdLineFile string) error {
+
+	if cmdLineFile == "" {
+		return fmt.Errorf("Kernel cmdline file cannot be empty")
+	}
+
+	kernelCmdline, err := ioutil.ReadFile(cmdLineFile)
+	if err != nil {
+		return err
+	}
+
+	words := strings.Fields(string(kernelCmdline))
+	for _, w := range words {
+		word := string(w)
+		if err := c.parseCmdlineOption(word); err != nil {
+			agentLog.Warnf("Failed to parse kernel option %s: %v", word, err)
+		}
+	}
+	return nil
+}
+
+func applyConfig(config agentConfig) {
+	agentLog.SetLevel(config.logLevel)
+}
+
+//Parse a string that represents a kernel cmdline option
+func (c *agentConfig) parseCmdlineOption(option string) error {
+
+	const (
+		optionPosition = iota
+		valuePosition
+		optionSeparator = "="
+	)
+
+	split := strings.Split(option, optionSeparator)
+
+	if len(split) < valuePosition+1 {
+		return nil
+	}
+
+	switch split[optionPosition] {
+	case logLevelFlag:
+		level, err := logrus.ParseLevel(split[valuePosition])
+		if err != nil {
+			return err
+		}
+		c.logLevel = level
+	default:
+		if strings.HasPrefix(split[optionPosition], optionPrefix) {
+			return fmt.Errorf("Unknown option %s", split[optionPosition])
+		}
+	}
+	return nil
+
 }
--- a/agent_test.go
+++ b/agent_test.go
@@ -20,6 +20,9 @@ import (
 	"io/ioutil"
 	"os"
 	"testing"
+
+	_ "github.com/opencontainers/runc/libcontainer/nsenter"
+	"github.com/sirupsen/logrus"
 )
 
 const testDisabledAsNonRoot = "Test disabled as requires root privileges"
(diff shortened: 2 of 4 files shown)


agent.go
```diff
@@ -1210,3 +1217,42 @@ func winsizeCb(pod *pod, data []byte) error {
...
+	words := strings.Fields(cmdline)
+	for _, w := range words {
+		if strings.HasPrefix(string(w), "agent.debug=") {
+			logLevel = strings.Split(string(w), "agent.debug=")[1]
+			break
+		}
+	}
+
+	if logLevel == "" {
+		agentLog.SetLevel(logrus.InfoLevel)
```
sboeuf: You should use `defaultLogLevel` that you defined earlier.

agent.go
```diff
@@ -1210,3 +1217,42 @@ func winsizeCb(pod *pod, data []byte) error {
...
+		return nil
+	}
+
+	level, err := logrus.ParseLevel(logLevel)
+	if err != nil {
+		agentLog.SetLevel(logrus.InfoLevel)
```
sboeuf: Ditto.

agent.go
```diff
@@ -221,7 +221,7 @@ func (p *pod) controlLoop(wg *sync.WaitGroup) {
 				continue
 			}
 
-			agentLog.Infof("Read on ctl channel failed: %v", err)
+			agentLog.Errorf("Read on ctl channel failed: %v", err)
```
sboeuf, line 241: Well that's not an error, it's about the proxy disconnecting, in order to end the agent.

network.go
```diff
@@ -137,7 +137,7 @@ func setupInterfaces(netHandle *netlink.Handle, ifaces []hyper.NetIface) error {
 			return fmt.Errorf("Interface HwAddr and Name are both empty")
 		}
 
-		agentLog.Infof("Link found %+v", link)
+		agentLog.Warnf("Link found %+v", link)
```
sboeuf: That's an info, not a warning.

time.go
```diff
@@ -0,0 +1,14 @@
+package main
```
sboeuf: Add the Intel license header please.

sboeuf (changes requested):

agent.go
jcvenegas: I think the only way the proxy disconnecting is doing DESTROYPOD ?

agent_test.go
```diff
@@ -39,3 +41,38 @@ func TestMain(m *testing.M) {
...
+		wantErr bool
+	}{
+		{"Empty cmdline",
+			args{""}, false},
+		{"Log config info",
+			args{"agent.debug=info"}, false},
+		{"Log config debug",
+			args{"agent.debug=debug "}, false},
+		{"Log config warn",
+			args{" systemd kernel=someting agent.debug=warn "}, false},
```
sboeuf: s/kernel=someting/kernel=foo/

agent_test.go
```diff
@@ -39,3 +41,38 @@ func TestMain(m *testing.M) {
...
+		{"Log config empty with space",
+			args{"agent.debug= debug "}, false},
+		{"Log config empty",
+			args{"agent.debug="}, false},
+	}
```
sboeuf: You could also test in case a wrong parameter is given. Something like:
`agent.debug=ino`
`agent.debug=wran`
`agent.debug=degbu`

agent_test.go
```diff
@@ -39,3 +41,38 @@ func TestMain(m *testing.M) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := parseKernelCmdline(tt.args.cmdline); (err != nil) != tt.wantErr {
+				t.Errorf("parseKernelCmdline() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			agentLog.Infof("info message")
+			agentLog.Debugf("debug message")
```
sboeuf: Those messages don't help in the test validation, do they ?
I think we need something a bit more complex. You could create a pipe with `os.Pipe()`, and give one end of the pipe to logrus, and listen on the other end. That way for instance, you could check that you receive `info message` only when you enable at least Info level.

agent.go
sboeuf: Ok fine.

agent.go
```diff
@@ -1241,13 +1241,13 @@ func parseKernelCmdline(cmdline string) error {
 	}
 
 	if logLevel == "" {
-		agentLog.SetLevel(logrus.InfoLevel)
+		agentLog.SetLevel(defaultLogLevel)
```
sboeuf: Addressed in the wrong commit.

agent.go
```diff
@@ -1241,13 +1241,13 @@ func parseKernelCmdline(cmdline string) error {
...
 		return nil
 	}
 
 	level, err := logrus.ParseLevel(logLevel)
 	if err != nil {
-		agentLog.SetLevel(logrus.InfoLevel)
+		agentLog.SetLevel(defaultLogLevel)
```
sboeuf: Ditto.

agent.go
```diff
@@ -171,6 +171,13 @@ func main() {
 	}
 
 	agentLog.Infof("Agent version: %s", Version)
+
+	if uptime, err := newEventTime(agentStartedEvent); err != nil {
+		agentLog.Errorf("Failed to get uptime %v", err)
```
sboeuf: Use a constant for `"Failed to get uptime"` because you reuse it later.

agent.go
```diff
@@ -799,7 +806,20 @@ func (p *pod) runCmd(cmd hyper.HyperCmd, data []byte) error {
 		return fmt.Errorf("No callback found for command %q", hyper.CmdToString(cmd))
 	}
 
-	return cb(p, data)
+	if t, err := newEventTime(hyper.CmdToString(cmd) + "_start"); err == nil {
+		agentLog.Infof("%s", t)
+	} else {
+		agentLog.Errorf("Failed to get time:%v", err)
```
sboeuf: Use a constant for the string.

agent.go
```diff
@@ -799,7 +806,20 @@ func (p *pod) runCmd(cmd hyper.HyperCmd, data []byte) error {
...
+	}
+
+	cbErr := cb(p, data)
+
+	if t, err := newEventTime(hyper.CmdToString(cmd) + "_end"); err == nil {
+		agentLog.Infof("%s", t)
+	} else {
+		agentLog.Errorf("Failed to get time:%v", err)
```
sboeuf: Ditto

time_test.go
```diff
@@ -0,0 +1,57 @@
+package main
```
sboeuf: Add the Intel license header please.

sboeuf (changes requested):

time.go
```diff
@@ -0,0 +1,60 @@
...
+// Get eventTime based CLOCK_MONOTONIC_RAW clock
+func newEventTime(event string) (eventTime, error) {
+	var ts unix.Timespec
+
+	if event == "" {
+		return eventTime{Event: "error", Time: ts}, fmt.Errorf("event cannot be empty")
+	}
+
+	if err := unix.ClockGettime(unix.CLOCK_MONOTONIC_RAW, &ts); err != nil {
+		return eventTime{Event: "error", Time: ts}, fmt.Errorf("%v", err)
```
jodh-intel: This looks odd to me for 2 reasons:

- conventially, if a function return an `error`, you should ignore any other returned values, but you are creating a special "error" event.
- why can't the returned error be just `, err` rather than `, fmt.Errorf("%v", err)` ?

agent.go
```diff
@@ -792,7 +806,20 @@ func (p *pod) runCmd(cmd hyper.HyperCmd, data []byte) error {
 		return fmt.Errorf("No callback found for command %q", hyper.CmdToString(cmd))
 	}
 
-	return cb(p, data)
+	if t, err := newEventTime(hyper.CmdToString(cmd) + "_start"); err == nil {
+		agentLog.Infof("%s", t)
+	} else {
+		agentLog.Errorf("Failed to get time:%v", err)
```
jodh-intel: It looks like the returned `error` won't include the event passed to `newEventTime()`. I'd create the string first:

```
event := hyper.CmdToString(cmd) + "_start"

if t, err := newEventTime(event); err == nil {
    agentLog.Infof("%s", t)
} else {
    agentLog.Errorf("Failed to get time for event %v: %v", event, err)
```

agent.go
```diff
@@ -1210,3 +1237,42 @@ func winsizeCb(pod *pod, data []byte) error {
...
+	}
+	return nil
+}
+
+func parseKernelCmdline(cmdline string) error {
+	logLevel := ""
+	words := strings.Fields(cmdline)
+	for _, w := range words {
+		if strings.HasPrefix(string(w), "agent.debug=") {
+			logLevel = strings.Split(string(w), "agent.debug=")[1]
```
jodh-intel: I'd create a variable rather than calling `string()` twice:

```
str := string(w)
```

Also, I think you should check the number of fields returned by `Split()` before de-referencing the 2nd element.

time.go
jcvenegas: you are right!, fixed

agent_test.go
jcvenegas: Indeed, it does not help, I added to help me know if was working but agree I can add the Pipe sugguestion

agent.go
```diff
@@ -1210,3 +1241,47 @@ func winsizeCb(pod *pod, data []byte) error {
 
 	return nil
 }
+
+// Get the agent configuration from kenrel cmdline
```
MarioCarrilloA: s/kenrel/kernel

agent.go
```diff
@@ -59,6 +59,9 @@ const (
 	statelessPassword  = "/usr/share/defaults/etc/passwd"
```
sboeuf: Please edit the commit message and replace: `s/kenel/kernel/`

agent.go
```diff
@@ -1210,3 +1222,51 @@ func winsizeCb(pod *pod, data []byte) error {
...
+//Parse a string that represents a kernel cmdline to get agent options
+func parseKernelCmdline(cmdline string) error {
+
+	const valuePosition = 1
+	const optionSeparator = "="
+
+	words := strings.Fields(cmdline)
+
+	for _, w := range words {
+		for o := range agentOptions {
```
sboeuf: I am sorry but I don't like that we loop on agentOptions and at the same time we are modifying its values. Try something  more simple to stay as clear as possible:
I think we should expect `getConfig()` to return a config, and for each possible option parsed by `parseKernelCmdLine()`, we want a default in case it is not provided by the command line.
This means that `parseKernelCmdLine()` will build a list of options, that is returned to `getConfig()`, and from the `main()`, we apply this list of options in any case because we know that we will always have some default values (a function like `applyConfig()` would be fine). No need to specify any error, maybe just a `log.Warn()` in case we could read `/proc/cmdline` for example.

Moreover, about the way you loop over every element of this, because this is a map, you can directly have `key/value` from the loop for:
```go
for key, value := range optionList {
    ...
}
```

agent.go
```diff
@@ -1261,7 +1261,7 @@ func parseKernelCmdline(cmdline string) error {
 
 	level, err := logrus.ParseLevel(agentOptions[logLevelFlag])
 	if err != nil {
-		agentLog.SetLevel(logrus.InfoLevel)
+		agentLog.SetLevel(defaultLogLevel)
```
sboeuf: This should be part of the previous commit.

agent.go
```diff
@@ -62,6 +62,7 @@ const (
 	kernelCmdlineFile  = "/proc/cmdline"
```
sboeuf: This commit message needs to be edited. `s/path/patch/` and `s/add/adds/`

sboeuf (changes requested):

time.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"golang.org/x/sys/unix"
+)
+
+const agentStartedEvent = "agent_started"
+
+//Event that will be reported for time tracing
+type eventTime struct {
+	Event string
+	Time  unix.Timespec
```
grahamwhaley, line 28: Hi @jcvenegas You asked me what time format I thought we should print in. My gut feeling is that we should probably use the same format as the other logs we produce within CC - so, for instance, it looks like the runtime logs use the logrus internal Time data:
https://github.com/clearcontainers/runtime/blob/9be2f54966b90014251afa5cebdf1ee36a85503f/logger.go#L137
If we can, I think we should do the same - what do you think @jodh-intel

jodh-intel: The actual timestamp itself should be in https://en.wikipedia.org/wiki/ISO_8601 format, or the RFC 3339 variant used by https://golang.org/pkg/time/. Looks like logrus uses that by default.

agent.go
jcvenegas: Hi @sboeuf , I updated the PR:
1. I created agentConfig struct (by now we only have logLevel)
2. getConfig() return parse the kernel cmdline 
 - If could not parse an option I will use a default option.
3. applyConfig(config) will take a config and apply the values already parsed and validated in getConfig()

agent.go
```diff
@@ -1210,3 +1221,59 @@ func winsizeCb(pod *pod, data []byte) error {
...
+		optionSeparator = "="
+	)
+
+	split := strings.Split(option, optionSeparator)
+
+	if len(split) < valuePosition+1 {
+		return nil
+	}
+
+	option = split[optionPosition]
```
sboeuf: Don't assign the key that you parsed to the option, instead, use the key (that is `split[optionPosition]`) directly into the switch:
```go
switch split[optionPosition] {
    ...
```

agent.go
```diff
@@ -1210,3 +1221,59 @@ func winsizeCb(pod *pod, data []byte) error {
...
+	value := split[valuePosition]
+
+	switch option {
+	case logLevelFlag:
+		level, err := logrus.ParseLevel(value)
+		if err != nil {
+			return err
+		}
+		c.logLevel = level
```
sboeuf: You could add a default case here, and return an error like:
```go
default:
    return fmt.Errorf("Unknown option %s", split[optionPosition])
```

sboeuf (commented):
Two comments, but other than that, the PR looks good to me now :)

agent.go
```diff
@@ -1210,3 +1244,59 @@ func winsizeCb(pod *pod, data []byte) error {
 
 	return nil
 }
+
+//Get the agent configuration from kernel cmdline
+func getConfig() agentConfig {
+
+	config := agentConfig{}
+	config.logLevel = defaultLogLevel
```
jodh-intel: I think it would be better if this function also returned an `error`. The caller can then set `config.logLevel = defaultLogLevel`. It's neat to encapsulate the default handling like this, but it makes testing it difficult (how can you know which codepath was followed if `kernelCmdlineFile` contained the same `defaultLogLevel` value?)

Speaking of which, can you add a unit test for this function :smile:

sboeuf (approved):

sboeuf:
LGTM
Let's wait for @jodh-intel comments before to merge it.

jcvenegas:
@jodh-intel I still need to rework the timestamp format...

jcvenegas:
@jodh-intel plase take a look to the log using logrus timestamp:
```
{"level":"info","msg":"Agent version: 0.1.0-alpha.0+-28cad9a9947ceced9a05b43ac3db5f4358ea5167","time":"2017-08-15T14:43:11.690803064Z"}
{"level":"info","msg":"agent_started 0.542392826","time":"2017-08-15T14:43:11.692726511Z"}
{"level":"info","msg":"Received \"startpod\" command","time":"2017-08-15T14:43:11.751076463Z"}
{"level":"info","msg":"startpod_start","time":"2017-08-15T14:43:11.751832257Z"}
{"level":"info","msg":"startpod_end","time":"2017-08-15T14:43:11.759136667Z"}
{"level":"info","msg":"Received \"newcontainer\" command","time":"2017-08-15T14:43:11.767398068Z"}
{"level":"info","msg":"newcontainer_start","time":"2017-08-15T14:43:11.767548997Z"}
```

jodh-intel:
@jcvenegas - I'm not clear on your last comment. This looks fine to me.

However, are you planning to either modify the logrus formatter to change the time format, or maybe add a special time format *message* for `startpod_start` and `startpod_end` or something?

jcvenegas:
@jodh-intel @grahamwhaley  I would like to get your approval in the format, now that I moved to RFC3339 as was suggested. 
- Yes I changed logrus time format to : `TimestampFormat: time.RFC3339Nano` to get nanoseconds precision.
- To log start end events I just print  `{event}_start` or  `{event}_end`
- An special case is logged `agent_started seconds.nanoseconds` this get uptime with good clock precision.  
- Logrus format is JSON for easy parsing

sboeuf:
@jodh-intel @grahamwhaley could you give us your review on this one ?

jodh-intel:
lgtm.

[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/70/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

grahamwhaley:
I'm happy if @jodh-intel is happy, and he is - so merging....

PR #72, opened 2017-08-08
agent: Reap child processes during the pod lifecycle

### LINKED ISSUE #71: Reap all the child processes
We have some issues with long term testing due to processes that are not properly reap. The consequence is that they remain as defunct (zombie) processes, and they retain some resources. We have to reap all those processes properly if we want those 'soak' tests to pass properly.

sboeuf:
All processes, whatever if they come from a new container or to new
processes started by an exec command, are children of cc-agent daemon
running inside the VM. This means that all of them have to be reap
by the parent process after they have terminated.

This patch defines the parent process as the new subreaper and it
starts a go routine aiming to reap any child process after it has
terminated. It does rely on the SIGCHLD signal to determine when
it can expect to reap one of his children.

Fixes #71

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"os/signal"
 	"path/filepath"
 	"runtime"
 	"strings"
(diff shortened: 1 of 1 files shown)


sboeuf:
Replaced by https://github.com/clearcontainers/agent/pull/73

PR #74, opened 2017-08-08
Wait for hotplugged device

amshinde:


syscall.go
```diff
@@ -111,6 +114,62 @@ func unmountShareDir() error {
...
+					break
+				}
+			}
+		}
+	}()
+
+	select {
+	case <-doneListening:
+		return
+	case <-time.After(1 * time.Second):
```
jcvenegas: Could you create a constant for the device timeout ?

amshinde: @jcvenegas I dont see the need to create a constant for this, I am using this value just once.

Gopkg.lock
```diff
@@ -63,6 +63,18 @@
 
 [[projects]]
   branch = "master"
+  name = "github.com/jkeiser/iter"
+  packages = ["."]
+  revision = "67b94d6149c6b2efe0fde15d36a50a7692c2da17"
+
+[[projects]]
+  branch = "master"
+  name = "github.com/jochenvg/go-udev"
```
sboeuf, line 72: Please make sure we force the constraint to the master branch in the `Gopkg.toml` for this repo because we import it on purpose and we want to stick the latest version.
About `"github.com/jkeiser/iter"`, I guess this is a dependency pulled by the new package we have imported. That's okay not to constraint this one because we are not using it directly.

Always make sure you run
```
dep ensure
dep prune
```
after you have modified `Gopkg.toml`

.ci/setup.sh
```diff
@@ -41,6 +41,9 @@ pushd "${runtime_repo_dir}"
 sed -i 's/cc-agent.target/clear-containers.target/' create.go
 popd
 
+# Install libudev-dev required for go-udev vendor dependency
+sudo apt-get install libudev-dev
```
sboeuf: This commit can be removed since this is installed from the `tests` repo. https://github.com/clearcontainers/tests/pull/337 just got merged.

syscall.go
sboeuf: You should remove `1 *` since you want 1 second.

I think we could wait up to 3 seconds, and in this case, the code should be:
```go
case <-time.After(time.Duration(3) * time.Second):
```
to avoid linter errors.

syscall.go
```diff
@@ -111,6 +114,62 @@ func unmountShareDir() error {
 	return os.RemoveAll(containerMountDest)
 }
 
+func waitForBlockDevice(deviceName string) (err error) {
+	devicePath := filepath.Join(devPath, deviceName)
+
+	if _, err := os.Stat(devicePath); err == nil {
```
sboeuf: Are we sure that because the device is here, it will be ready ?
I am asking because this would mean that we could call into `Stat()` from an infinite loop, until we either find the path, or hit the timeout.

syscall.go
```diff
@@ -111,6 +114,62 @@ func unmountShareDir() error {
...
+
+	// Create a monitor listening on a NetLink socket
+	monitor := u.NewMonitorFromNetlink("udev")
+
+	// Add filter to watch for just block devices
+	if err := monitor.FilterAddMatchSubsystemDevtype("block", "disk"); err != nil {
+		return err
+	}
+
+	// Create done signal channel for signalling epoll loop on the monitor socket.
```
sboeuf, line 134: s/signalling/signaling/

syscall.go
```diff
@@ -111,6 +114,62 @@ func unmountShareDir() error {
...
+	ch, _ := monitor.DeviceChan(done)
+
+	go func() {
+		agentLog.Infof("Started listening for udev events for block device hotplug")
+
+		// Check if the device already exists
+		if _, err := os.Stat(devicePath); err == nil {
+			agentLog.Infof("Device %s already hotplugged, quit listening", deviceName)
+			close(done)
+			close(doneListening)
```
sboeuf: Please put a return at the end of this statement, this would allow avoiding the `else` statement which is not needed.

syscall.go
```diff
@@ -111,6 +114,62 @@ func unmountShareDir() error {
...
+		} else {
+			for d := range ch {
+				agentLog.Infof("Event:", d.Syspath(), d.Action())
+				if d.Action() == "add" && filepath.Base(d.Devpath()) == deviceName {
+					agentLog.Infof("Hotplug event received for device %s", deviceName)
+					close(doneListening)
```
sboeuf: Why are you not closing `done` channel here ?

sboeuf (changes requested):
Few changes needed. And also need clarification about `Stat()` vs `netlink`. I guess that's because Stat() would be a polling loop, and netlink is more like an interrupt, right ?

syscall.go
grahamwhaley: @amshinde sometimes a constant is just nice for making it clearer to the reader what has been chosen/set and why - it doesn't have to hinge on the 'multi use' case.
Sure, it is a balance act, and in this case this is a pretty small file. I don't think it is a blocker.

syscall.go
grahamwhaley: I guess that is true. If we do go that route, obviously put a small nap/sleep in the loop so we don't hard spin on it.

amshinde: Yes, successful stat on the device would mean the device is ready. I went with the route of udev events as I find it a much cleaner approach, rather than polling continuously.

syscall.go
amshinde: @sboeuf Duration of 1 second is more than enough, if the device is not available by then means something has gone wrong.

.ci/setup.sh
sboeuf: Nevermind, this commit is completely fine, and we have to remove the one from the `tests` repo since it does not build the agent.

syscall.go
amshinde: Tend to follow British English :)

sboeuf: Oh! my bad...

amshinde:
@sboeuf Changes pushed.

syscall.go
sboeuf: I definitely agree it is cleaner, thanks for the confirmation.

syscall.go
```diff
@@ -111,6 +114,61 @@ func unmountShareDir() error {
...
+			}
+		}
+		close(doneListening)
+	}()
+
+	select {
+	case <-doneListening:
+		close(done)
+	case <-time.After(time.Duration(1) * time.Second):
+		close(done)
```
sboeuf, line 166: Please error in this case.

Gopkg.toml
```diff
@@ -93,3 +93,7 @@
 [[constraint]]
   branch = "master"
   name = "golang.org/x/sys"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/jochenvg/go-udev"
```
sboeuf, line 99: Great, thanks !

sboeuf (commented):
Only one comment about returning an error in case of timeout. Other than that this PR sounds ready to be merged.

syscall.go
amshinde: Done

sboeuf (approved):

PR #80, opened 2017-08-17
Protect against race conditions

amshinde:


sboeuf:
@amshinde I'll take a look at this today

sboeuf:
@amshinde actually I have noticed the `wip` in the title, let me know when it is ready so that I can review !

amshinde:
@sboeuf PR is ready for review

devimc:
lgtm

[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/80/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent) [Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/80/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

amshinde:
Looks like a couple of tests are timing out while killing a container in semaphore. Will be able to take a look at this tomorrow.
@sboeuf  let me know what you think about the changes.

agent.go
```diff
@@ -120,15 +120,16 @@ type container struct {
...
+	id             string
+	running        bool
+	containers     map[string]*container
+	ctl            *os.File
+	tty            *os.File
+	stdinList      map[uint64]*os.File
+	network        hyper.Network
+	stdinLock      sync.Mutex
+	ttyLock        sync.Mutex
+	containersLock sync.RWMutex
```
sboeuf, line 144: I would call this one `podLock` instead of `containersLock` since it actually locks the Pod.

agent.go
```diff
@@ -197,6 +198,24 @@ func main() {
...
+
+	return ctr
+}
+
+func (p *pod) setContainer(ID string, ctr *container) {
+	p.containersLock.Lock()
+	p.containers[ID] = ctr
+	p.containersLock.Unlock()
+}
+
```
sboeuf: You should create a new function `deleteContainer()` locking the removal of a container:
```go
func (p *pod) deleteContainer(id string) {
    p.containersLock.Lock()
    delete(p.containers, id)
    p.containersLock.Unlock()
}
```
And use it from the code below

agent.go
```diff
@@ -1096,11 +1124,13 @@ func removeContainerCb(pod *pod, data []byte) error {
 		return fmt.Errorf("Container %s running, impossible to remove", payload.ID)
 	}
 
-	if err := pod.containers[payload.ID].removeContainer(payload.ID); err != nil {
+	if err := ctr.removeContainer(payload.ID); err != nil {
 		return err
 	}
 
+	pod.containersLock.Lock()
```
sboeuf: This won't be enough here. Mutexes are a first step, but we also needs some synchronization waiting for all the started processes to end before we try to remove the container. Basically, this means we need to wait on a `sync` before to perform the container's removal here. Every time we create/start a new process, we should do something like `wg.Add(1)` and every time it ends, call into `wg.Done()`. This way we ensure we are not removing a container which is still used.

agent.go
```diff
@@ -117,6 +117,7 @@ type container struct {
 	container libcontainer.Container
 	config    configs.Config
 	processes map[string]*process
+	pod       *pod
```
sboeuf: Refactoring sounds okay to me, but you should give the pod pointer through the parameters of the function `closeProcessStreams()` since you need the pod pointer only in this function. This would avoid to add this pointer in every container.

Moreover, please fix your commit message, there are some syntax/grammar errors.

agent.go
```diff
@@ -114,10 +114,11 @@ type process struct {
...
 type container struct {
-	container libcontainer.Container
-	config    configs.Config
-	processes map[string]*process
-	pod       *pod
+	container     libcontainer.Container
+	config        configs.Config
+	processes     map[string]*process
+	pod           *pod
+	processesLock sync.RWMutex
```
sboeuf, line 130: I would call this one `containerLock` instead of `processesLock` since it locks the entire container.

agent.go
```diff
@@ -565,67 +566,99 @@ func (p *pod) sendSeq(seq uint64, data []byte) error {
...
+
+	return proc
+}
+
+func (c *container) setProcess(pid string, process *process) {
+	c.processesLock.Lock()
+	c.processes[pid] = process
+	c.processesLock.Unlock()
+}
+
```
sboeuf, line 617: You could add the same type of function for the process removal:
```go
func (c *container) deleteProcess(pid string) {
    ...
}
```

agent.go
```diff
@@ -197,6 +198,24 @@ func main() {
 	wgLoops.Wait()
 }
 
+func (p *pod) getContainer(ID string) (ctr *container) {
```
sboeuf: `ID` should be lower case `id`.

agent.go
```diff
@@ -197,6 +198,24 @@ func main() {
...
+
+	ctr, exist := p.containers[ID]
+	if exist == false {
+		return nil
+	}
+
+	return ctr
+}
+
+func (p *pod) setContainer(ID string, ctr *container) {
```
sboeuf: `ID` should be lower case `id`.

sboeuf (changes requested):

agent.go
amshinde: @sboeuf This only protects the processes map, hence the name.

agent.go
amshinde: @sboeuf This only protects the containers map, so I have used the name containersLock here.

agent.go
amshinde: done

amshinde:
@sboeuf  I have pushed changes, have a look

sboeuf:
Thanks @amshinde

agent.go
sboeuf: Fair enough.

agent.go
sboeuf: This comment has not been addressed.

agent.go
sboeuf: Great, this has been added.

agent.go
sboeuf: Fair enough.

agent.go
```diff
@@ -114,10 +114,11 @@ type process struct {
 }
 
 type container struct {
```
sboeuf, line 125: Please rework the commit message, there are some syntax/grammar errors.

agent.go
sboeuf: Great, you have addressed this one.

agent.go
```diff
@@ -1103,6 +1105,8 @@ func newContainerCb(pod *pod, data []byte) error {
 
 	pod.setContainer(payload.ID, container)
 
+	container.wgProcesses.Add(1)
```
sboeuf: This can be moved to `runContainerProcess()` in order to avoid calling this both from `newContainerCb()` and `execCb()`. There is no risk to increase the waitgroup counter from this go routine as long as we make sure the counter is incremented before the signal is sent through the `started` Go channel.

agent.go
```diff
@@ -1229,6 +1235,8 @@ func execCb(pod *pod, data []byte) error {
 
 	ctr.setProcess(payload.Process.ID, process)
 
+	ctr.wgProcesses.Add(1)
```
sboeuf: Ditto. (can be moved)

sboeuf (changes requested):
Few comments

agent.go
amshinde: @sboeuf iirc Add should execute before the statement creating the goroutine and not in the goroutine,  hence the two places.

sboeuf: Oh... You think it does not work even if we have the handler `container.wgProcesses` from the goroutine ?
I thought it would be okay since the starting of any process (newContainer or exec) waits for the `started` channel for synchronization.

amshinde:
@sboeuf Commit messages have been fixed

sboeuf (approved):

sboeuf:
LGTM

PR #87, opened 2017-08-22
network: Don't try to replace initial existing routes

### LINKED ISSUE #86: Fix failing routes setup
The routes setup was failing after we have been replacing the initial route by the same one but without giving the source. Indeed, this initial route is setup when the interface is added, and it does not need to be replaced/overridden by the one provided through the `startpod` command. We skip this entry from the provided route list, allowing all the following routes to be properly added/replaced.

sboeuf:
Everytime an IP address is created for a specific interface, a route
is created accordingly. Because this route is a reference, we don't
want to modify it when applying the route list provided by the
startpod command. Instead, we ignore this specific entry from the
route list.

Fixes #86

### DIFF
--- a/network.go
+++ b/network.go
@@ -148,9 +148,25 @@ func setupInterfaces(netHandle *netlink.Handle, ifaces []hyper.NetIface) error {
 }
 
 func setupRoutes(netHandle *netlink.Handle, routes []hyper.Route) error {
+	initRouteList, err := netHandle.RouteList(nil, netlink.FAMILY_ALL)
+	if err != nil {
+		return err
+	}
+
 	for _, route := range routes {
 		var dst *net.IPNet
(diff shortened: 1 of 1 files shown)


sboeuf:
Also fixes https://github.com/clearcontainers/runtime/issues/432

sboeuf:
@amshinde please take a look.

egernst (commented):
lgtm

Thanks sboeuf

amshinde (approved):
lgtm

egernst:
lgtm

[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/87/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

PR #90, opened 2017-08-28
agent: implement ps command

### LINKED ISSUE #91: implement ps command
https://github.com/clearcontainers/runtime/issues/95

### LINKED ISSUE #95: Implement the runtime ps command
`ps` shows all processes running inside a container.

devimc:
ps command is used to list the processes running inside
the container, this command is called by ```docker top```

fixes #91
partially fixes https://github.com/clearcontainers/runtime/issues/95

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"bytes"
 	"encoding/binary"
 	"encoding/json"
 	"fmt"
@@ -26,6 +27,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
 	"syscall"
@@ -172,6 +174,12 @@ var callbackList = map[hyper.HyperCmd]cmdCb{
 	hyper.WinsizeCmd:         winsizeCb,
 }
 
+type cmdCbWithReply func(*pod, []byte) ([]byte, error)
+
+var callbackWithReplyList = map[hyper.HyperCmd]cmdCbWithReply{
+	hyper.PsContainerCmd: processListCb,
+}
+
 var agentLog = logrus.WithFields(logrus.Fields{
 	"name": name,
 	"pid":  os.Getpid(),
@@ -297,12 +305,17 @@ func (p *pod) controlLoop(wg *sync.WaitGroup) {
 
 		fieldLogger.Info("Received command")
 
-		if err := p.runCmd(cmd, data); err != nil {
+		response, err := p.runCmd(cmd, data)
+		if err != nil {
 			fieldLogger.WithError(err).Info("command failed")
 			reply = hyper.ErrorCmd
 		}
 
-		if err := p.sendCmd(reply, []byte{}); err != nil {
+		if response == nil {
+			response = []byte{}
+		}
+
+		if err := p.sendCmd(reply, response); err != nil {
 			fieldLogger.WithError(err).Info("reply send failed")
 			break
 		}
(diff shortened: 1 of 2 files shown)


agent.go
```diff
@@ -1172,50 +1181,155 @@ func (c *container) removeContainer(id string) error {
...
+
+		// appends pid line
+		for _, pid := range pids {
+			if pid == p {
+				result.WriteString(line + "\n")
+				break
+			}
+		}
+	}
+
```
sboeuf, line 1372: I get that this whole function is about duplicating what `runc` does here https://github.com/opencontainers/runc/blob/master/ps.go#L41-L95, but I think we should add an added value, that is gather all those data inside a structure dedicated to this. This structure would be then marshalled so that it would become a JSON byte array that we can send to the runtime(virtcontainers basically).

agent.go
```diff
@@ -156,6 +159,7 @@ var callbackList = map[hyper.HyperCmd]cmdCb{
 	hyper.ReadyCmd:           readyCb,
 	hyper.PingCmd:            pingCb,
 	hyper.WinsizeCmd:         winsizeCb,
+	hyper.PsContainerCmd:     processListCb,
```
sboeuf: Because it has to return a reply, it should be part of a different map. Create a new one called `callbackWithReplyList`

agent.go
```diff
@@ -892,33 +901,33 @@ func (p *pod) buildProcess(hyperProcess hyper.Process) (*process, error) {
 	return p.buildProcessWithoutTerminal(proc)
 }
 
-func (p *pod) runCmd(cmd hyper.HyperCmd, data []byte) error {
+func (p *pod) runCmd(cmd hyper.HyperCmd, data []byte) ([]byte, error) {
 	cb, exist := callbackList[cmd]
 	if exist == false {
```
sboeuf: This check should not return an error anymore. Indeed, relying on what I have mentioned above, if we add this new Hyper command to a new map called `callbackWithReplyList`, we should not fail but instead, we should try this other list. If we cannot find for both lists, then we can really fail this time.
This would allow not to force all the other commands to return nil all the time because they have no byte to send back.

sboeuf (changes requested):

agent.go
devimc: do you mean to put all output inside a struct?

```
type psOutput struct {
    output byte[]
}
```

I wonder if this is really needed

sboeuf: @devimc no, I meant having something nicer than only an array of byte already in the expected format. Basically, I was thinking that the agent could parse the output of the `ps` command to put it into a structure like:
```go
type processInfo struct {
        Name string
        PID     string
        ...
}

type processListInfo []processInfo
```
And this would be marshalled into a json byte array through the response. Then we would have virtcontainers relying on this specific structure to unmarshal the byte array. And then, this structure would be passed back to the runtime. The runtime wrapper being the only one doing the formatting.

I am saying all of this because if we think about virtcontainers as a generic library, we want to return a process list as a standardized structure that any consumer of virtcontainers could format the way it wants.

devimc: @sboeuf  the problem with the structure that you propose is that we don't know the options user will provide

sboeuf: That's why we need to have every potential output covered by this structure. The output will depend on the options, I agree, but eventually we would be sending the same structure.

devimc: the list of possible formats/columns is huge, I think  that is not a good idea. I wouldn't like to update/maintain this structure each time a new column/format is added/removed in the ```ps``` command

sboeuf: @devimc Yes I understand...
@sameo can we get your input on this please ?

devimc:
@sameo ping

agent.go
sameo: @devimc I'd agree with @sboeuf here. The agent should send data without assuming anything about the end user expected format. The runtime should be the one doing the formatting/filtering, not the agent.

devimc: @sameo great! , so can we merge?

sameo: @devimc I don't think you're marshalling an actual structure that the runtime/virtcontainers could unmarshal and format filter, right ?

devimc: ```runc``` doesn't format/filter the output because docker does that [1], I consider we should have a similar implementation and leave to docker decide the output format.

[1] - https://github.com/moby/moby/blob/402540708c9a0c35dc0b279a0f330455633537b8/daemon/top_unix.go#L63

sameo: All right, let's get this one in.

sameo:
LGTM

[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/90/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent) [Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/90/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

sameo:
@sboeuf Please let us know if you're ready to LGTM that PR.

sboeuf:
LGTM

[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/90/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent) [Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/90/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

sboeuf (approved):

PR #131, opened 2017-10-12
kill: fix sporadic deadlock related to freezer cgroup issue

### LINKED ISSUE #130: kill: Deadlock waiting for a kill to be completed
In few cases, we end up waiting forever for our container to be signalled. This issue comes from a libcontainer issue here: https://github.com/opencontainers/runc/issues/1609 that has been solved with https://github.com/opencontainers/runc/pull/1610
We need to re-vendor our agent in order to get the fix.

sboeuf:
In few cases, we end up waiting forever for our container to be signalled. This issue comes from a libcontainer issue here: opencontainers/runc#1609 that has been solved with opencontainers/runc#1610
We need to re-vendor our agent in order to get the fix.

Fixes #130


agent.go
```diff
@@ -1213,9 +1213,19 @@ func killContainerCb(pod *pod, data []byte) error {
...
+		// Use AllProcesses to make sure we carry forward the flag passed by the runtime.
+		signalled <- ctr.container.Signal(payload.Signal, payload.AllProcesses)
+	}()
+
+	select {
+	case err := <-signalled:
+		if err != nil {
+			return fmt.Errorf("Process could not be signalled: %v", err)
+		}
+	case <-time.After(time.Duration(15) * time.Second):
```
grahamwhaley, line 1227: I wonder if we should relate this 15 to the docker '23s' timeout (10+10+3) I think we saw during the bug hunt - just a thought.

grahamwhaley (approved):
One thought, otherwise
lgtm

agent.go
sboeuf: Well I thought about it and that's why I wanted a timeout shorter so that we can actually get the logs printed before docker completely force kill everything after this 23s timeout.

PR #139, opened 2017-10-20
agent: re-scan pci bus to discover hidden devices

### LINKED ISSUE #138: Add support for Q35 machines
re-scan pci bus before create a container
qemu <= 2.9 does not support ACPI hot plug for Q35 machines hence pci bus must be re-scanned manually

devimc:
qemu <= 2.9 does not support ACPI pci hotplug in Q35 machines,
this means the linux kernel will not receive the order to
re-enumerate/re-scan the pci devices connected to the buses,
hence pci bus must be re-scanned manually looking for hidden devices

fixes #138

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -65,6 +65,8 @@ const (
 	logLevelFlag       = optionPrefix + "log"
 	defaultLogLevel    = logrus.InfoLevel
 	cannotGetTimeMsg   = "Failed to get time for event %s:%v"
+	pciBusRescanFile   = "/sys/bus/pci/rescan"
+	pciBusMode         = 0220
 )
 
 var capsList = []string{
@@ -1059,6 +1061,12 @@ func addMounts(config *configs.Config, fsmaps []hyper.Fsmap) error {
 
 func newContainerCb(pod *pod, data []byte) error {
 	var payload hyper.NewContainer
+
+	// re-scan PCI bus
+	// looking for hidden devices
+	if err := ioutil.WriteFile(pciBusRescanFile, []byte("1"), pciBusMode); err != nil {
+		agentLog.WithError(err).Warnf("Could not rescan PCI bus")
+	}
 
 	if pod.running == false {
 		return fmt.Errorf("Pod not started, impossible to run a new container")


agent.go
```diff
@@ -1060,6 +1062,12 @@ func addMounts(config *configs.Config, fsmaps []hyper.Fsmap) error {
 func newContainerCb(pod *pod, data []byte) error {
 	var payload hyper.NewContainer
 
+	// re-scan PCI bus
+	// looking for hidden devices
+	if err := ioutil.WriteFile(pciBusRescanFile, []byte("1"), pciBusMode); err != nil {
+		agentLog.WithError(err).Warnf("Could not open pci bus rescan file %s: %s", pciBusRescanFile, err)
```
sboeuf: `"Could not rescan the PCI bus"` would be more generic and more appropriate. The underlying reason why this is happening will come from `err`.

sboeuf (commented):
This looks good (I have only one comment about the log), but I have a question regarding the commit message.
If I read the commit message, I have the feeling that Qemu <= 2.9 cannot support ACPI hotplug, meaning that we cannot hotplug anything. And at the same time, you're saying that we have to rescan the PCI bus from inside the VM in that case. Could you elaborate how you can still "hotplug" things (without ACPI), and that you need to rescan because they don't show up automatically. I am confused and it would be great if the commit message was more verbose to explain this change (I trust you that the fix is really needed but I'd like to understand exactly why).

devimc:
@sboeuf done

sboeuf (approved):
Thanks @devimc 
This looks better now ;)

sboeuf:
LGTM

agent.go
```diff
@@ -1060,6 +1062,12 @@ func addMounts(config *configs.Config, fsmaps []hyper.Fsmap) error {
 func newContainerCb(pod *pod, data []byte) error {
 	var payload hyper.NewContainer
 
+	// re-scan PCI bus
+	// looking for hidden devices
+	if err := ioutil.WriteFile(pciBusRescanFile, []byte("1"), pciBusMode); err != nil {
+		agentLog.WithError(err).Warnf("Could not rescan pci bus: %s", err)
```
jodh-intel: This is going to log the error twice, so I'd make this just:

```
agentLog.WithError(err).Warn("Could not rescan PCI bus")
```

grahamwhaley (changes requested):
lgtm
but let's fix that log message @jodh-intel noted.

devimc:
@grahamwhaley  @jodh-intel  done, thanks

jodh-intel:
lgtm

[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/139/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

grahamwhaley (approved):
lgtm

agent.go
```diff
@@ -1060,6 +1062,12 @@ func addMounts(config *configs.Config, fsmaps []hyper.Fsmap) error {
 func newContainerCb(pod *pod, data []byte) error {
 	var payload hyper.NewContainer
 
+	// re-scan PCI bus
+	// looking for hidden devices
+	if err := ioutil.WriteFile(pciBusRescanFile, []byte("1"), pciBusMode); err != nil {
```
mcastelino, line 1067: @devimc In the other PR you are reusing the pci slots. Not sure if a slot will ever be reused (block device unplug followed by a plug in the real world with virtcontainers). However if that does happen, what happens to the mount points?

Also how do we know when the rescan is complete?

devimc: @mcastelino pci slots can be reused once it's free

> what happens to the mount points?

There is still missing that part, in qemu <= 2.9 Q35 PCI devices must be removed manually, I guess we have to implement an extra command in the agent to clean up unplugged devices

> Also how do we know when the rescan is complete?

it is a blocking IO operation, process will not continue until all PCI buses are re-scanned

sboeuf:
@mcastelino what is the status on this ? Are you okay with this PR, or you expect some changes ?

mcastelino:
LGTM. We do not remount block devices now. So this is fine.

[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/139/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent) [Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/139/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

agent.go
mcastelino: @devimc will the pci slot reuse happen in the real world. Do we support removing a container from a POD?

devimc: not sure if we can remove a container from a POD, but we can unplug devices from bridges.
we could reuse that slot to hot plug another device if we need it since we have a limitation in the number of devices per bridge.

mcastelino: @devimc so when will we ever unplug a device? I cannot think of a case. Which means we are ok for now and need no special logic or unmount/remount.

PR #143, opened 2017-10-31
agent: Discard signals if cannot find sequence ID

### LINKED ISSUE #142: Discard signals if cannot find sequence ID or container is stopped
When a container with terminal enabled is started, it is very likely that it's going to receive a SIGWINCH signal right after it started. But there is an issue here, if the workload exited very quickly, when the signal arrives to the agent, it cannot find the process/seqID (because it has been removed when the process exited).

The race condition happens because at the same time, the agent sends the exit code to the shim, and the shim sends the signal to the agent.

The way to avoid such issues is to proceed the same way it was done with the sdtin coming from IO channel, that is dropping the signal if we have no seqID match. That's why we should discard signals, either because there is no corresponding seqID, or because the container holding those seqID has been stopped. In both cases, we cannot prevent the agent to receive signals, and the best we can do
is to ignore them.

sboeuf:
When a container with terminal enabled is started, it is very likely
that it's going to receive a SIGWINCH signal right after it started.
But there is an issue here, if the workload exited very quickly, when
the signal arrives to the agent, it cannot find the process/seqID.

The race condition happens because at the same time, the agent sends
the exit code to the shim, and the shim sends the signal to the agent.

The way to avoid such issues is to proceed the same way it was done
with the sdtin coming from IO channel, that is dropping the signal
if we have no seqID match. That's why this patch discards signals,
either because there is no corresponding seqID, or because the
container holding those seqID has been stopped. In both cases, we
cannot prevent the agent to receive signals, and the best we can do
is to ignore them.

Fixes #142

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -1226,9 +1226,17 @@ func killContainerCb(pod *pod, data []byte) error {
 	}
 
 	ctr := pod.getContainer(payload.ID)
-
 	if ctr == nil {
 		return fmt.Errorf("Container %s not found, impossible to signal", payload.ID)
+	}
+
+	status, err := ctr.container.Status()
+	if err != nil {
+		return err
+	}
+	if status == libcontainer.Stopped {
+		agentLog.Info("Container %s is Stopped on pod %s, discard signal %s", payload.ID, pod.id, payload.Signal.String())
+		return nil
 	}
 
 	signalled := make(chan error)
@@ -1362,7 +1370,7 @@ func pingCb(pod *pod, data []byte) error {
 	return nil
 }
 
-func (p *pod) findTermFromSeqID(seqID uint64) (*os.File, string, error) {
+func (p *pod) findTermFromSeqID(seqID uint64) (*os.File, string) {
 	p.containersLock.RLock()
 	defer p.containersLock.RUnlock()
 
@@ -1372,13 +1380,13 @@ func (p *pod) findTermFromSeqID(seqID uint64) (*os.File, string, error) {
 		for _, process := range container.processes {
 			if process.seqStdio == seqID {
 				container.processesLock.RUnlock()
-				return process.termMaster, cid, nil
+				return process.termMaster, cid
 			}
 		}
 		container.processesLock.RUnlock()
 	}
 
-	return nil, "", fmt.Errorf("Could not find a process related to sequence %d", seqID)
+	return nil, ""
 }
 
 func winsizeCb(pod *pod, data []byte) error {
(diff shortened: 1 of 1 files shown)


agent.go
```diff
@@ -1392,20 +1400,26 @@ func winsizeCb(pod *pod, data []byte) error {
...
 
-	term, cid, err := pod.findTermFromSeqID(payload.Sequence)
-	if err != nil {
-		return err
+	term, cid := pod.findTermFromSeqID(payload.Sequence)
+	if cid == "" {
+		// The sequence ID could be not found in case the process returned
+		// and the process has been removed from the container map.
+		// We should not error in that case, and discard the received signal.
+		agentLog.Warnf("Could not find sequence ID %d, discard SIGWINCH", payload.Sequence)
```
jodh-intel: Again, could be useful to have the pod ID logged here.

agent.go
```diff
@@ -1226,11 +1226,19 @@ func killContainerCb(pod *pod, data []byte) error {
...
 	if ctr == nil {
 		return fmt.Errorf("Container %s not found, impossible to signal", payload.ID)
 	}
 
+	status, err := ctr.container.Status()
+	if err != nil {
+		return err
+	}
+	if status == libcontainer.Stopped {
+		agentLog.Info("Container not running anymore, discard signal %s", payload.Signal.String())
```
jodh-intel: Please can you add the container ID to this message to help with debugging.

agent.go
```diff
@@ -1392,20 +1400,26 @@ func winsizeCb(pod *pod, data []byte) error {
...
 	status, err := ctr.container.Status()
-
 	if err != nil {
 		return err
 	}
 
-	if status != libcontainer.Running {
-		return fmt.Errorf("Container %s not running, impossible to resize window", cid)
+	if status == libcontainer.Stopped {
+		agentLog.Info("Container not running anymore, discard SIGWINCH")
```
jodh-intel: Pod+Container ID here?

agent.go
grahamwhaley: and if you are there, substitute 'not running' for 'is Stopped' - may as well use the official status state to help us decode the logs later.

grahamwhaley (commented):
Good catch.
I'm presuming we saw evidence of the race - maybe just some reported failures in the logs or some error about bad exit codes or something - might be good to note that in the commit or pull message.
With the nits about log messages - l g t m

sboeuf:
@grahamwhaley well I don't have logs from this particular failure since I have discovered this while I was debugging some other issues and by running a simple docker command in a loop, I have realized the issue because I was getting some sporadic failures from the agent. The commit message seems enough verbose to understand what I am fixing here, don't you think ?

sboeuf:
@grahamwhaley @jodh-intel I have addressed your comments.

jodh-intel:
lgtm

[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/143/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

grahamwhaley (approved):
@sboeuf sure, if we don't have the error logs, np. It's nice practice to include them if we can - it helps folks later on who run into the same issues as they can search more effectively to see if they have been found/fixed before.

lgtm

Issue #144, opened 2017-11-01
docker (master) kill and docker kill KILL return exit code 255


jcvenegas:
The current docker master branch in docker now calls the runtime to signal a container. Instead of send a signal directly to the workload process (in our case cc-shim).

So now the container workload insided a clear container gets a KILL signal but the agent returns the exit code is 255 and not 137.
Example:
```bash
docker run --runtime cc-runtime --name bC7Bdk3yLYGjTty2EBHyv4rTZAKMU6 -ti busybox sh -c 'trap "exit 15" 15 ; while : ; do sleep 1; echo wait; done'
sleep 5
docker kill -s 9 bC7Bdk3yLYGjTty2EBHyv4rTZAKMU6
sleep 5
# This will be 255 but is expected to be 137
docker inspect --format={{.State.ExitCode}} bC7Bdk3yLYGjTty2EBHyv4rTZAKMU6
```

PR #145, opened 2017-11-01
exit: Check if process was signaled

jcvenegas:
If the process was signaled with SIGTERM the exit
code reported was 255, but the correct should be 137.

This patch uses the issue based in the suggestion
in a golang issue[1]. If the process exited because was
signaled we return 128+signal.

Fixes: #144

[1] https://github.com/golang/go/issues/7938

Signed-off-by: Jose Carlos Venegas Munoz <jose.carlos.venegas.munoz@intel.com>

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -67,6 +67,10 @@ const (
 	cannotGetTimeMsg   = "Failed to get time for event %s:%v"
 	pciBusRescanFile   = "/sys/bus/pci/rescan"
 	pciBusMode         = 0220
+	// If a process terminates because of signal "n"
+	// The exit code is "128 + signal_number"
+	// http://tldp.org/LDP/abs/html/exitcodes.html
+	exitSigalOffset = 128
 )
 
 var capsList = []string{
@@ -870,9 +874,19 @@ func (p *pod) runContainerProcess(cid, pid string, terminal bool, started chan e
 		fieldLogger.Info("Got process state")
 
 		if waitStatus, ok := processState.Sys().(syscall.WaitStatus); ok {
-			exitCode = uint8(waitStatus.ExitStatus())
-			fieldLogger.WithField("exit-code", exitCode).Info("got wait exit code")
+			exitStatus := waitStatus.ExitStatus()
+
+			if waitStatus.Signaled() {
+				exitCode = uint8(exitSigalOffset + waitStatus.Signal())
+				fieldLogger.WithField("exit-code", exitCode).Info("process was signaled")
+			} else {
+				exitCode = uint8(exitStatus)
+				fieldLogger.WithField("exit-code", exitCode).Info("got wait exit code")
+			}
 		}
+
+	} else {
+		fieldLogger.Error("Process state is nil could not get process exit code")
 	}
 
 	// Send exit code through tty channel


sboeuf:
@jcvenegas This change looks good to me, but are you sure `processState` is not `nil` in case of `SIGKILL` ? I mean have you tested this change and does that work for you with containerd 1.0 ?

jodh-intel:
lgtm.

Nit: shouldn't the commit mention `143` (`128 + 15` [`SIGTERM`]) rather than `137` (which would be for `SIGKILL`)?

[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/145/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

grahamwhaley (commented):
Nice pickup

jodh-intel:
Random thought: now that we're using structured logging, in principle we could modify `kill_test.go` to not only check the return code, but also to ensure that -- when debug is enabled -- the correct signal/exit code is logged.

agent.go
```diff
@@ -870,9 +870,19 @@ func (p *pod) runContainerProcess(cid, pid string, terminal bool, started chan e
 		fieldLogger.Info("Got process state")
 
 		if waitStatus, ok := processState.Sys().(syscall.WaitStatus); ok {
-			exitCode = uint8(waitStatus.ExitStatus())
-			fieldLogger.WithField("exit-code", exitCode).Info("got wait exit code")
+			exitStatus := waitStatus.ExitStatus()
+
+			if waitStatus.Signaled() {
+				exitCode = uint8(128 + waitStatus.Signal())
```
devimc: could you please define a variable and document why 128 ?

jcvenegas: @devimc changed to constant and documented :D

jcvenegas:
@sboeuf  ` but are you sure processState is not nil in case of SIGKILL ? I mean have you tested this change and does that work for you with containerd 1.0 ?`

I double check and it is not nil - it returns a `255` exit code, googling I found that issue, by checking if exited by a signal seems to work well with all the cases (contained 1.0 and latest docker release).

agent.go
```diff
@@ -67,6 +67,10 @@ const (
 	cannotGetTimeMsg   = "Failed to get time for event %s:%v"
 	pciBusRescanFile   = "/sys/bus/pci/rescan"
 	pciBusMode         = 0220
+	// If a process finish because fatal error with  signal "n"
```
sboeuf: s/finish/terminates/
s/because fatal error with signal/because of signal/

sboeuf:
@jcvenegas that's great if you checked ! I like the way this is handled, pretty neat!
LGTM

[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/145/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent) [Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/145/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

agent.go
jcvenegas: Changed :)

sboeuf (approved):

PR #151, opened 2017-11-06
agent-cli: Allow reconnection to the agent

### LINKED ISSUE #150: agent-cli: Allow reconnection to the agent
Because the agent sends a READY message the first time a connection is established, there is no way to reconnect without killing/restarting the agent. In order to workaround this, a new flag is added to the agent CLI so that we can make a difference between those two cases.

sboeuf:
Because the agent sends a READY message the first time a connection is established, there is no way to reconnect without killing/restarting the agent. In order to workaround this, a new flag is added to the agent CLI so that we can make a difference between those two cases.

Fixes #150

### DIFF
--- a/hack/agent-cli/README.md
+++ b/hack/agent-cli/README.md
@@ -32,8 +32,10 @@ cc-agent
 
 From a different shell on the host (as root):
 ```
-# ./agent-cli run --ctl=/tmp/hyper.sock --tty=/tmp/tty.sock
+# ./agent-cli run --ctl=/tmp/hyper.sock --tty=/tmp/tty.sock --wait-for-ready
 ```
+
+Do not use the `--wait-for-ready` flag when the `agent-cli` is terminated or disconnected from the `CTL` and `TTY` sockets and the agent is still listening on those sockets. In this case, the agent will not send a `READY` message.
 
 ## Usage
 
--- a/hack/agent-cli/agent_cli.go
+++ b/hack/agent-cli/agent_cli.go
@@ -233,20 +233,21 @@ func monitorStdInLoop(h *hyperstart.Hyperstart, done chan<- bool) error {
 		}
 	}
 
-	done <- true
+	close(done)
+
 	return nil
 }
 
 func monitorTtyOutLoop(h *hyperstart.Hyperstart, done chan<- bool) error {
 	for {
-		msgCh := make(chan *hyperstart.TtyMessage, 1)
-		errorCh := make(chan bool, 1)
+		msgCh := make(chan *hyperstart.TtyMessage)
+		errorCh := make(chan bool)
 
 		go func() {
 			msg, err := h.ReadIoMessage()
 			if err != nil {
 				magicLog("%s\n", err)
-				errorCh <- true
+				close(errorCh)
 				return
 			}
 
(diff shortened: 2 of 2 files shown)


grahamwhaley:
Hi @sboeuf . It's not clear to me from the PR when this would be used. In my head the agent will still be running in the container/VM when we try to reconnect to it - so when do we get to issue the extra command line parameter? I suspect I'm missing something in the whole 'flow' here of reconnection - any chance you can clarify that in the PR message?

Oh, and the CIs failed due to `check`. I wonder if that is related to your https://github.com/clearcontainers/agent/pull/153 at all?

grahamwhaley:
```
22:20:30 hack/agent-cli/agent_cli.go:262:4:warning: ineffective break statement. Did you mean to break out of the outer loop? (SA4011) (megacheck)
22:20:30 syscall.go:38:5:warning: should omit comparison to bool constant, can be simplified to recursive (S1002) (megacheck)
22:20:30 agent.go:1493:17:warning: unnecessary conversion (unconvert)
22:20:30 Makefile:67: recipe for target 'check' failed
22:20:30 make: *** [check] Error 1
22:20:30 Build step 'Execute shell' marked build as failure
```

sboeuf:
@grahamwhaley yes the CI is failing because of our discussion on #153. About this patch, it only affects the agent-cli tool, so no worries I am not modifying the agent behavior here. The reason for this is that we have some folks from the security team playing with this tool and they needed to be able to reconnect after a CTRL-C of this CLI tool. The problem is that our agent sends a READY message after someone connects to it, but only the first time. This PR is kind of a WA in order to reconnect without expecting a READY message.

grahamwhaley:
ah, thanks - that explains it.
I'll get to fixing #153 in a bit...

sboeuf:
@grahamwhaley thanks ;)

sboeuf:
@grahamwhaley @iphutch can you please take a look ?

hack/agent-cli/README.md
```diff
@@ -32,9 +32,11 @@ cc-agent
 
 From a different shell on the host (as root):
 ```
-# ./agent-cli run --ctl=/tmp/hyper.sock --tty=/tmp/tty.sock
+# ./agent-cli run --ctl=/tmp/hyper.sock --tty=/tmp/tty.sock --wait-for-ready
 ```
 
+In case the agent-cli is terminated or disconnected from `CTL` and `TTY` sockets, and the agent is still listening on those sockets, you should not use the flag `--wait-for-ready` since the agent will not send any `READY` message.
```
iphutch: I suggest a note:

note:
Do not use the `--wait-for-ready` flag when the agent-cli is terminated or disconnected from the `CTL` and `TTY` sockets and the agent is still listening on those sockets. In this case, the agent will not send a `READY` message.

sboeuf: Thanks for the comment, this is better. I have updated the code.

iphutch (approved):
lgtm

jcvenegas:
lgtm

jcvenegas (approved):

sboeuf:
@jcvenegas could you give lgtm once again ?

sboeuf:
@jcvenegas and merge it after that please.

jcvenegas (approved):

jcvenegas:
lgtm

jcvenegas:
lgtm

grahamwhaley (approved):
I wonder if 'wait-for-ready' should be the default on or off - but, am believing this works for the intended use case, so:
lgtm

PR #158, opened 2017-11-10
network: Update route list according to what's on the system

### LINKED ISSUE #157: network: Update route list according to what's on the system
We need to update the list of routes based on the ones already on the system.

sboeuf:
When using the destroyPod() function, the removal of the network is
performed but it generates an error. The error says that we cannot
remove the route. This is caused by the fact that we should not try
to remove this route since we have not added it. Indeed, when we do
the network setup, we skip the existing routes.

This patch removes routes already on the system from the route list
held in the pod structure. That way, when destroyPod() is called,
the list does not contain the routes that we have not added.

Fixes #157

### DIFF
--- a/network.go
+++ b/network.go
@@ -154,13 +154,19 @@ func setupInterfaces(netHandle *netlink.Handle, ifaces []hyper.NetIface) error {
 	return nil
 }
 
-func setupRoutes(netHandle *netlink.Handle, routes []hyper.Route) error {
+func setupRoutes(netHandle *netlink.Handle, routes *[]hyper.Route) error {
 	initRouteList, err := netHandle.RouteList(nil, netlink.FAMILY_ALL)
 	if err != nil {
 		return err
 	}
 
-	for _, route := range routes {
+	var finalRouteList []hyper.Route
+	defer func() {
+		// Update the route list.
+		*routes = finalRouteList
+	}()
+
+	for _, route := range *routes {
 		var dst *net.IPNet
 		var err error
 
@@ -209,6 +215,11 @@ func setupRoutes(netHandle *netlink.Handle, routes []hyper.Route) error {
 			return fmt.Errorf("Could not add/replace route dest(%s)/src(%s)/gw(%s)/dev(%s): %v",
 				route.Dest, route.Src, route.Gateway, route.Device, err)
 		}
+
+		// Only save the routes that we are actually adding. Don't add
+		// skipped routes already existing, otherwise it could cause
+		// issues when trying to remove them.
+		finalRouteList = append(finalRouteList, route)
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


network.go
```diff
@@ -198,6 +199,11 @@ func setupRoutes(netHandle *netlink.Handle, routes []hyper.Route) error {
 			}
 		}
 
+		// Only save the routes that we are actually adding. Don't add
+		// skipped routes already existing, otherwise it could cause
+		// issues when trying to remove them.
+		finalRouteList = append(finalRouteList, route)
```
jcvenegas: dumb questions, you are returning `finalRouteList` only in the case  of a successful setup. But later if `removeRoutes` is called to remove (clean up) the added interfaces. It will try to remove all of them and not only the ones that were added.

So questions:
1. It is need to return the `finalRouteList` even on fail ?
2. In case being need to return finalRouteList or just to have according to the comment (`Only save the routes that we are actually adding`), can you move `finalRouteList = append(finalRouteList, route)` after call `netHandle.RouteReplace`

sboeuf: Good points, I am gonna make some changes.

sboeuf:
@jcvenegas please take a look, I've addressed your comments !

jcvenegas:
cri-o test failed due to pull timeout , rerunning 
```
00:15:08.093 time="2017-11-10T23:48:27Z" level=error msg="error copying docker://redis:alpine to dir:/home/jenkins/jenkins_slave/workspace/clear-containers-agent-fedora-26-PR/go/src/github.com/kubernetes-incubator/cri-o/.artifacts/redis-image: Error reading blob sha256:b56ae66c29370df48e7377c8f9baa744a3958058a766793f821dadcb144a4647: Get https://dseasb33srnrn.cloudfront.net/registry-v2/docker/registry/v2/blobs/sha256/b5/b56ae66c29370df48e7377c8f9baa744a3958058a766793f821dadcb144a4647/data?Expires=1510358897&Signature=Y~ekRZMp8ua-R1bPnUbgU5iWtFCn3-io~oJmmZZxCN3c0wrhOF3ycPD~murScjGiyfLVG5inKRkhaGQOx38pUkSSPz54qUw8TvzL3nw1haIq6ZAW0nu8BZ4348naG4rS5wqADqZdMdNBASFaDLo0J8wje56CNKk6ZvkBUYteyJI_&Key-Pair-Id=APKAJECH5M7VWIS5YZ6Q: dial tcp: lookup dseasb33srnrn.cloudfront.net on 168.63.129.16:53: read udp 10.0.0.6:42380->168.63.129.16:53: i/o timeout" 
```

jcvenegas (approved):
lgtm

network.go
jcvenegas: looks good, thanks !

jcvenegas:
@sboeuf  looks like goveralls is  bit sensitive with the decreased code coverage

sboeuf:
@jcvenegas yeah I guess we have to change this, but that's not a big deal, we can merge it anyway. Just waiting for the CI

jodh-intel:
lgtm

[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/158/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

PR #160, opened 2017-11-10
destroy: Change DESTROYPOD behavior not to exit

### LINKED ISSUE #159: Change DESTROYPOD behavior by not making the agent exit
DESTROYPOD command should not be responsible for making the agent process exit and terminate the VM as a consequence. We need the VM to be killed on purpose by the runtime and not to be implicitely killed from the agent.

sboeuf:
This patch is simple but it modifies the complete behavior of the
DESTROYPOD command since it won't cause the agent process to exit
and it won't cause the VM to exit subsequently.

Indeed, we expect the runtime to be responsible for destroying the
VM, and not the agent itself. Moreover, this behavior could allow
the support for a restart of the pod without restarting the VM.
The future will tell us if that's something we need to implement.

Fixes #159

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -306,10 +306,6 @@ func (p *pod) controlLoop(wg *sync.WaitGroup) {
 			fieldLogger.WithError(err).Info("reply send failed")
 			break
 		}
-
-		if cmd == hyper.DestroyPodCmd {
-			break
-		}
 	}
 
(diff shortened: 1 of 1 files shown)


grahamwhaley (approved):
As you say, a simple change with big consequences :-)
lgtm

PR #162, opened 2017-11-13
stdin: Close process stdin when receiving EOF

### LINKED ISSUE #161: Detect EOF of stdin on TTY channel
Receiving an empty message through the tty channel (I mean empty message but with a valid sequence ID) should be treated as an EOF for the stdin of the corresponding process. The agent should close stdin in that case.

sboeuf:
In case the shim stdin is closed, it is going to receive an empty
message when reading from it. This behavior has to be reproduced
on the agent side so that it can determine when to close stdin.

The proxy will forward an empty message and this patch will make
the agent aware that receiving an empty message for a valid seq
ID should be processed as an EOF. In that case, stdin will be
closed and the process will end up properly (if it was waiting
on the input to finish).

Fixes #161

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -151,11 +151,16 @@ type pod struct {
 	containers     map[string]*container
 	ctl            *os.File
 	tty            *os.File
-	stdinList      map[uint64]*os.File
+	stdinList      map[uint64]stdinInfo
 	network        hyper.Network
 	stdinLock      sync.Mutex
 	ttyLock        sync.Mutex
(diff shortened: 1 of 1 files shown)


jodh-intel (approved):

sboeuf:
CI needs https://github.com/clearcontainers/tests/pull/719 to run without failures.

sboeuf:
@jcvenegas could you take a look.

grahamwhaley (approved):
lgtm

PR #170, opened 2017-11-17
Capabilities: Limit the list of capabilities

mcastelino:
Limit the list of capabilities to the same subset as runc

Fixes https://github.com/clearcontainers/runtime/issues/818

Signed-off-by: Manohar Castelino <manohar.r.castelino@intel.com>

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -121,6 +121,23 @@ var capsList = []string{
 	"CAP_WAKE_ALARM",
 }
 
+var defaultCapsList = []string{
+	"CAP_CHOWN",
+	"CAP_DAC_OVERRIDE",
+	"CAP_FOWNER",
+	"CAP_FSETID",
+	"CAP_KILL",
+	"CAP_SETGID",
+	"CAP_SETUID",
+	"CAP_SETPCAP",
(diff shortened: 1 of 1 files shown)


sboeuf:
The code is fine, but I am wondering if that's what we want cause right now we don't have support for adding capabilities and I always thought that this change (your PR) should come along with the support of adding a list of capabilities from the config.json provided.
I am worried that with this change, some commands will not work anymore because we will ignore what they could provide/need as capabilities to be added.
@sameo what do you think about it ?

mcastelino:
@sboeuf it should not change anything. We run unpriv containers. So we are actually providing CAPs that should not have been provided in the first place.

sboeuf:
LGTM

[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/170/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent) [Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/170/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

sboeuf (approved):

PR #174, opened 2017-11-28
agent: add CAP_SYS_ADMIN capability

### LINKED ISSUE #173: Unable to mount block device inside the container
```sh
$ docker run -e http_proxy=$http_proxy --device /dev/loop1p1 -ti debian bash
root@53b62083b411:/# mount /dev/loop1p1 /mnt/
mount: permission denied
```

container must have ```CAP_SYS_ADMIN``` capability

devimc:
container must be able to mount/umount block device, for this
reason CAP_SYS_ADMIN must be part of the default capabilities list

fixes #173

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -135,6 +135,7 @@ var defaultCapsList = []string{
 	"CAP_MKNOD",
 	"CAP_AUDIT_WRITE",
 	"CAP_SETFCAP",
+	"CAP_SYS_ADMIN",
 }
 
 type agentConfig struct {


sboeuf:
The change works but the real fix would be to implement the whole chain inside runtime + virtcontainers + agent. What I mean is that we need to get this capability from the config.json provided by Docker.
@mcastelino recently reduced the list of default CAPS to the same used in runc, and I don't think we want such a PR to be merged right now. @mcastelino please comment on this one.

devimc:
@sboeuf ohh ya, make sense
@sboeuf  @mcastelino  feel free to close the PR

PR #178, opened 2017-12-05
agent: Make the agent the subreaper of all processes

### LINKED ISSUE #177: Make the agent process as subreaper
Look at the following case:
```
Container process ==spawn==> child process ==spawn==> granchild process
```
If the child terminates before the grandchild process, then the grandchild is going to be reparented to the container process (init/pid1 of the PID namespace). The problem with this case is that our container process is not a reaper, it won't wait for all its descendants like a reaper would do.
That's why we should make the agent as the subreaper of all its descendants.

sboeuf:
agent: Make the agent the subreaper of all processes
    
    Our agent had a flaw, it potentially lives with some zombie processes
    according to what has been done by the container user. The explanation
    is that we only wait for container and exec processes but we don't
    handle the children and grandchildren processes coming from those
    initial container and exec processes. The example of such a case can
    be shown demonstrated in case the container process spawn a child
    process, and this one will also spawn another child process. In case
    the child terminates before its own children, those ones expect to be
    reaped by the init process of the PID namespace, i.e the container
    process. But this container process is not written to reap descendant
    processes, which will lead to a zombie process left behind.
    
    This commit solves those cases by setting the agent process, who will
    be the father of all container processes, as a subreaper. This way, it
    will receive all the SIGCHLD signals from any child left behind, and
    will reap them. Notice this patch makes sure we don't check the error
    from the libcontainer call to process.Wait() since it won't be able to
    reap the container or exec processes as expected. This libcontainer
    Wait() function is only used here to properly cleanup the pipes and
    internal structures created by libcontainer.
    
    Fixes #177
    

sboeuf:
Fixes #177

sboeuf:
Cannot be merged because of libcontainer issue: https://github.com/opencontainers/runc/issues/1677

sboeuf:
opencontainers/runc#1677 has been fixed with opencontainers/runc#1678 which has been merged !
This PR can be unblocked now.

sboeuf:
@sameo @amshinde please take a look.

sboeuf:
@jodh-intel @grahamwhaley @jcvenegas @sameo @amshinde 
Please take a look :)

devimc:
I wonder if runC has the same issue

> In case the child terminates before its own children, those ones expect to be reaped by the init process of the PID namespace

yee, but 

> If the "init" process of a PID namespace terminates, the kernel terminates all of the processes in the namespace via a SIGKILL signal  (pid_namespaces(7))

basically once the workload ends, all processes are killed, but I guess you want to handle the case where the workload never ends and its children spawn and leave processes running

lgtm

[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/178/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent) [Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/178/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

sboeuf:
@devimc, `runc` does not have the same issue because it actually set itself as a subreaper (there a flag to disable the subreaper if needed).
And yes you're right, I am trying to handle the case where the container process does not end, but the workload has started a bunch of children and grandchildren who might terminate after their parent.

reaper.go
```diff
@@ -0,0 +1,191 @@
...
+
+	"github.com/opencontainers/runc/libcontainer"
+	"golang.org/x/sys/unix"
+)
+
+type reaper struct {
+	sync.RWMutex
+
+	chansLock     sync.RWMutex
+	exitCodeChans map[int]chan int
```
amshinde: Can you add a comment explaining this structure?

reaper.go
```diff
@@ -0,0 +1,191 @@
...
+		if pid < 1 {
+			return nil
+		}
+
+		status := exitStatus(ws)
+
+		agentLog.Infof("SIGCHLD pid %d, status %d", pid, status)
+
+		exitCodeCh, err := r.getExitCodeCh(pid)
+		if err != nil {
```
amshinde: Can you make it clear in the comments  that we are only interested in the exit code of the Container and Exec processes, and ignoring the exit code from children and grandchildren(just reaping them)

agent.go
```diff
@@ -856,12 +895,36 @@ func (p *pod) runContainerProcess(cid, pid string, terminal bool, started chan e
 
 	fieldLogger := agentLog.WithField("container-pid", pid)
 
+	// This lock is very important to avoid any race with reaper.reap().
+	// Indeed, if we don't lock this here, we could potentially get the
+	// SIGCHLD signal before the channel has been created, meaning we will
```
amshinde: Can you mention exit code channel to be more clear.

agent.go
```diff
@@ -896,44 +959,20 @@ func (p *pod) runContainerProcess(cid, pid string, terminal bool, started chan e
...
-	if _, ok := err.(*exec.ExitError); err != nil && !ok {
-		fieldLogger.WithError(err).Error("Process wait failed")
+	// Using helper function wait() to deal with the subreaper.
+	libContProcess := (*reaperLibcontainerProcess)(&(proc.process))
+	exitCode, err := p.subreaper.wait(processID, libContProcess)
+	if err != nil {
+		return err
 	}
-	// Close pipes to terminate routeOutput() go routines.
-	ctr.closeProcessPipes(pid)
```
amshinde, line 907: this is not required anymore?

sboeuf: This is still called from the defer here: https://github.com/clearcontainers/agent/pull/178/files/e8d8a5fc6359f172177a782b85f153011d008015..c36a44adc6e06215b1c4cb736aa1e5d57699f82f#diff-2fe34e1b6f7f9aece0af74b1635502eeR889

reaper.go
sboeuf: Sure I can.

reaper.go
sboeuf: Yes good point, I'll try to add more comments !

agent.go
sboeuf: Ok.

sboeuf:
@amshinde I have updated the PR with more comments.

amshinde (approved):
lgtm

sboeuf:
@amshinde Could you merge if you're fine with this ?

PR #180, opened 2017-12-09
Listen to udev "add" events to detect devices that have been added and bind mount these within pod's containers. 

amshinde:
From within a container, if a device is attached to a user driver, resulting in new device nodes being created under /dev, these are not visible from within a container. This is because /dev is mounted using tmpfs in a separate mount namespace. We need to continue using tmpfs, to restrict access to the initial set of devices available to the container. But we do need to provide a way for the container to access devices created later on based on a whitelist/blacklist.

This PR fixes this by watching for devices created and bind mounts these to each container's /dev so that the container is able to access the devices.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -156,14 +156,15 @@ type process struct {
 }
 
 type container struct {
-	container       libcontainer.Container
-	config          configs.Config
-	processes       map[string]*process
-	pod             *pod
-	processesLock   sync.RWMutex
-	wgProcesses     sync.WaitGroup
-	deviceNodes     []string
-	deviceNodesLock sync.Mutex
+	container        libcontainer.Container
+	config           configs.Config
+	processes        map[string]*process
+	pod              *pod
+	processesLock    sync.RWMutex
+	wgProcesses      sync.WaitGroup
+	deviceNodes      []string
+	deviceNodesLock  sync.Mutex
+	systemMountsInfo hyper.SystemMountsInfo
 }
 
 type pod struct {
@@ -559,8 +560,10 @@ func (p *pod) bindDeviceNode(devNode string) {
 			"ctr":     ctr.container.ID(),
 		}).Info("Bind mounting device to /dev")
 
-		if err := ctr.bindDeviceNode(devNode); err != nil {
-			fieldLogger.WithError(err).Info("Could not bind device")
+		if ctr.systemMountsInfo.BindMountDev {
+			if err := ctr.bindDeviceNode(devNode); err != nil {
+				fieldLogger.WithError(err).Info("Could not bind device")
+			}
 		}
 	}
 }
@@ -1341,12 +1344,6 @@ func newContainerCb(pod *pod, data []byte) error {
 				Device:      "sysfs",
 				Flags:       defaultMountFlags,
 			},
-			{
-				Source:      "/dev/vfio",
-				Destination: "/dev/vfio",
-				Device:      "bind",
-				Flags:       syscall.MS_BIND | syscall.MS_REC,
-			},
 		},
 
 		NoNewKeyring:    true,
@@ -1379,10 +1376,11 @@ func newContainerCb(pod *pod, data []byte) error {
 	processes[payload.Process.ID] = builtProcess
 
 	container := &container{
-		pod:       pod,
-		container: libContContainer,
-		config:    config,
-		processes: processes,
+		pod:              pod,
+		container:        libContContainer,
+		config:           config,
+		processes:        processes,
+		systemMountsInfo: payload.SystemMountsInfo,
 	}
 
 	pod.setContainer(payload.ID, container)
--- a/agent.go
+++ b/agent.go
@@ -35,6 +35,7 @@ import (
 	"unsafe"
 
 	hyper "github.com/clearcontainers/agent/api"
+	goudev "github.com/jochenvg/go-udev"
 	"github.com/opencontainers/runc/libcontainer"
 	"github.com/opencontainers/runc/libcontainer/configs"
 	_ "github.com/opencontainers/runc/libcontainer/nsenter"
@@ -155,12 +156,14 @@ type process struct {
 }
 
 type container struct {
-	container     libcontainer.Container
-	config        configs.Config
-	processes     map[string]*process
-	pod           *pod
-	processesLock sync.RWMutex
-	wgProcesses   sync.WaitGroup
+	container       libcontainer.Container
+	config          configs.Config
+	processes       map[string]*process
+	pod             *pod
+	processesLock   sync.RWMutex
+	wgProcesses     sync.WaitGroup
+	deviceNodes     []string
+	deviceNodesLock sync.Mutex
 }
 
 type pod struct {
@@ -205,6 +208,20 @@ var agentLog = logrus.WithFields(logrus.Fields{
 	"name": name,
 	"pid":  os.Getpid(),
 })
+
+// Add kernel subsystems that need to be blacklisted here.
+// These devices will not be mounted within a container.
+var blacklistedDeviceSubsystems = []string{}
+
+func isBlacklistedSubsystem(subsystem string) bool {
+	for _, s := range blacklistedDeviceSubsystems {
+		if subsystem == s {
+			return true
+		}
+	}
+
+	return false
+}
 
 func init() {
 	if len(os.Args) > 1 && os.Args[1] == "init" {
@@ -268,7 +285,14 @@ func main() {
 	// Run TTY loop
 	go pod.streamsLoop(&wgLoops)
 
+	// Create done signal channel for signalling udev epoll loop.
+	done := make(chan struct{})
+
+	// Run loop listening for udev events
+	pod.listenToUdevEvents(done)
+
 	wgLoops.Wait()
+	close(done)
 }
 
 func (p *pod) getContainer(id string) (ctr *container) {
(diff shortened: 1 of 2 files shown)


jodh-intel:
CI is barfing on lint errors.

agent.go
```diff
@@ -480,6 +488,69 @@ func (p *pod) closeChannels() {
...
+				"udev-path":  d.Syspath(),
+				"udev-event": d.Action(),
+			}).Info("Received udev event")
+
+			// Ignore udev events for block devices, these are handled by another go-routine
+			if d.Subsystem() == "block" {
+				continue
+			}
+
+			// Add subsystems that need to be blacklisted here
```
jodh-intel: I'd add in code to support blacklisting now. The list might be empty, but it will then be easier to support future blacklisting, but also it will allow this code to be tested.

Speaking of tests, can you add some? If unit tests aren't possible/realistic, can you create a set of tests in https://github.com/clearcontainers/tests somewhere to exercise this new functionality?

amshinde: @jodh-intel See https://github.com/clearcontainers/tests/issues/794.

amshinde: I have added in the code to support blacklisting.

agent.go
```diff
@@ -480,6 +502,70 @@ func (p *pod) closeChannels() {
...
+				"udev-path":  d.Syspath(),
+				"udev-event": d.Action(),
+			}).Info("Received udev event")
+
+			// Ignore udev events for block devices, these are handled by another go-routine
+			if d.Subsystem() == "block" {
+				continue
+			}
+
+			if isBlacklistedSubsystem(d.Subsystem()) {
```
jodh-intel, line 538: Thanks! Just one other comment - it might be useful to add a log call when a device is being black-listed.

agent.go
```diff
@@ -480,6 +502,70 @@ func (p *pod) closeChannels() {
...
+
+		fieldLogger.Info("Started listening for all udev events")
+
+		for d := range ch {
+			fieldLogger.WithFields(logrus.Fields{
+				"udev-path":  d.Syspath(),
+				"udev-event": d.Action(),
+			}).Info("Received udev event")
+
+			// Ignore udev events for block devices, these are handled by another go-routine
```
sameo: Which routine handle block devices?

agent.go
```diff
@@ -480,6 +502,70 @@ func (p *pod) closeChannels() {
...
+				continue
+			}
+
+			if isBlacklistedSubsystem(d.Subsystem()) {
+				continue
+			}
+
+			if d.Action() == "add" {
+				// Bind device to /dev within all containers
+				p.bindDeviceNode(d.Devnode())
```
sameo, line 549: Should we unbind those devices when deleting the container?

agent.go
```diff
@@ -1161,6 +1161,11 @@ func newContainerCb(pod *pod, data []byte) error {
 		return err
 	}
 
+	dest := filepath.Join(absoluteRootFs, "/dev")
+	if err := syscall.Mount("tmpfs", dest, "tmpfs", syscall.MS_NOSUID|syscall.MS_STRICTATIME, ""); err != nil {
```
sameo: Same here, I think we should umount that one when deleting the container.

mcastelino: @amshinde @sboeuf should we not setup a /dev directory per container. Not all containers may be allowed to see all mount points. Maybe I am reading this wrong and this is what you are already doing.

agent.go
mcastelino: @amshinde why all the containers. Should it not be based on per container device assignment. At this point the device list and mounts come per container config.json right

/cc @sboeuf

agent.go
```diff
@@ -480,6 +504,80 @@ func (p *pod) closeChannels() {
 	}
 }
 
+// listenToUdevEvents listens to udev events to detect device nodes added
+// and binds the device nodes to the /dev of all the containers.
+// This is required as a container has its /dev mounted with tmpfs with an initial
+// list of devices.
```
mcastelino, line 511: It would be good to call out when this happens. I assume this is for the use case where devices are specified as part of the container config.json and you are hot plugging them into the VM, and now expect them to show up in the specific container?

mcastelino: Or is this due to a device creation due to some action in the container which has the CAPs to perform device creation, and this device is created due to an activity in the container itself does not show up in the container.

amshinde: @mcastelino It is the latter case. When you are performing some action in the container that results in a device node being created in the VM. This is useful for running DPDK inside the container where you would typically bind a device to vfio or uio resulting in a new device node being created.

agent.go
amshinde: When the device node is created with the user interacting on the container command line, it is not possible to know which container the device node should be created in. But since all containers in a pod are supposed to share the same namespaces for k8s, I think this behaviour should be ok.

agent.go
sboeuf: As discussed offline, this should be gated by the fact that the user has requested a mount `/dev:/dev`. This is the way we would do that with runc, and this is the proper way to make sure the user expects to see new devices being created and to be able to get access to them.

sboeuf: This raises the question of handling special cases like this one, where cannot expect to mount the directory from the host. Should we maintain a list of such directories ? Actually, we should define such a list first and see how we plan to handle every case.

agent.go
amshinde: @sameo I have done all the unbinding of devices including /dev. 
But I am seeing failures with CRIO.

agent.go
amshinde: @sboeuf See this commit where I have addressed `-v /dev:/dev` : https://github.com/clearcontainers/agent/pull/180/commits/0680f5a626c85ad5daf299e2bc1903b945ce1a5d

amshinde:
@sameo @sboeuf I pushed a few changes addressing `-v /dev:/dev` and unmounting of devices. Please have a look.

sboeuf:
thx @amshinde , I'll check that

amshinde:
@sboeuf I have introduced similar change in virtcontainers to pass information about the /dev bind mount:
https://github.com/containers/virtcontainers/pull/523

That needs to go in first.

agent.go
```diff
@@ -1131,6 +1131,21 @@ func addMounts(config *configs.Config, fsmaps []hyper.Fsmap) error {
 	return nil
 }
 
+func mountDevtoRootfs(absoluteRootFs string) error {
```
sboeuf: s/`mountDevtoRootfs`/`mountDevToRootfs`/

agent.go
```diff
@@ -1378,10 +1375,11 @@ func newContainerCb(pod *pod, data []byte) error {
...
 	container := &container{
-		pod:       pod,
-		container: libContContainer,
-		config:    config,
-		processes: processes,
+		pod:          pod,
+		container:    libContContainer,
+		config:       config,
+		processes:    processes,
+		bindMountDev: payload.BindMountDev,
```
sboeuf: The change looks okay to me, but same comment as for virtcontainers, this should be part of a more generic structure, so that container would have only one new field for any exceptions we have to make regarding bind mounts.

sboeuf (commented):
@amshinde about the "listen udev events" commit, why don't you merge the block device case with this case, that is having a global go routine listening constantly to uevents. You could provide a way to add more rules in order to make sure we don't miss any uevent you could expect.

amshinde:
@sboeuf I want to keep the two cases separate, as block devices are mounted based on the name passed on the command line.
```
docker run -it debian --device /dev/sdb:/dev/sdf.
```
Here the /dev/sdb passed to the VM may become "/dev/vda", we then need to mount that as "/dev/sdf" inside the container and not "/dev/vda". Hence we handle that separately.

sboeuf:
@amshinde ah yes that makes sense. There are so many cases to handle, I am worried about this complexity to handle many specific cases...

amshinde:
Yeah I know! Its just that many of the runc configs dont translate very well to us.

sboeuf:
@amshinde Yes agreed... Then please try to document into the code as much as you can so that it is easier to anyone who wants to take a look at it.

amshinde:
@sboeuf Will do.

amshinde:
@sboeuf I have included changes that I made to virtcontainers here

sboeuf:
LGTM

[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/180/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent) [Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/180/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

amshinde:
@sboeuf I want to vendor changes into the proxy before this goes in.

sboeuf:
Ok

amshinde:
@sboeuf I have vendored changes in proxy. Can you merge this if you are ok with this?

sboeuf (approved):

mcastelino (commented):
@amshinde can you clarify what you mean by

``We also get rid of the VFIO bind mount, which was an interim solution.``

amshinde:
@mcastelino before this PR, I had implemented a temporary solution of bind-mounting /dev/vfio directory within the container, so that devices bound to vfio can be seen.
With this PR we no longer require the temporary solution.

PR #187, opened 2017-12-23
WIP: scsi: Add support to pass SCSI address for SCSI disks

amshinde:
Add support for SCSI disk to be passed for container rootfs.
If SCSI address is provided, use that to get the SCSI disk name.
This eliminates the need to predict the drive name on the host side.
We do need to incur the cost of scanning SCSI bus.

Signed-off-by: Archana Shinde <archana.m.shinde@intel.com>

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -526,7 +526,9 @@ func (p *pod) listenToUdevEvents(done chan struct{}) {
 		for d := range ch {
 			fieldLogger.WithFields(logrus.Fields{
 				"udev-path":  d.Syspath(),
+				"dev-path":   d.Devpath(),
 				"udev-event": d.Action(),
+				"event":      d,
 			}).Info("Received udev event")
 
 			// Ignore udev events for block devices, these are handled by another go-routine in waitForBlockDevice
@@ -1283,10 +1285,13 @@ func newContainerCb(pod *pod, data []byte) error {
 		return fmt.Errorf("Pod not started, impossible to run a new container")
 	}
 
+	agentLog.WithField("xxxData", data).Info("Data before parsing")
+
 	if err := json.Unmarshal(data, &payload); err != nil {
 		return err
 	}
 
+	agentLog.WithField("xxxpayload", payload).Info("Payload parsed")
 	if payload.Process.ID == "" {
 		payload.Process.ID = fmt.Sprintf("%d", payload.Process.Stdio)
 	}
--- a/agent.go
+++ b/agent.go
@@ -1295,7 +1295,7 @@ func newContainerCb(pod *pod, data []byte) error {
 		return fmt.Errorf("Container %s already exists, impossible to create", payload.ID)
 	}
 
-	absoluteRootFs, err := mountContainerRootFs(payload.ID, payload.Image, payload.RootFs, payload.FsType)
+	absoluteRootFs, err := mountContainerRootFs(payload.ID, payload.Image, payload.RootFs, payload.FsType, payload.SCSIAddr)
 	if err != nil {
 		return err
 	}
--- a/api/commands.go
+++ b/api/commands.go
@@ -182,6 +182,9 @@ type NewContainer struct {
 	Fsmap            []Fsmap          `json:"fsmap"`
 	Process          Process          `json:"process"`
 	SystemMountsInfo SystemMountsInfo `json:"systemMountsInfo"`
+
+	// SCSI address in the format SCSI-Id:LUN
+	SCSIAddr string `json:"scsiAddr,omitempty"`
 }
 
 // KillContainer describes the format expected by a KILLCONTAINER command.
--- a/syscall.go
+++ b/syscall.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"syscall"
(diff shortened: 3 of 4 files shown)


amshinde:
Will be sending virtcontainer changes for SCSI support

agent.go
```diff
@@ -1158,7 +1158,7 @@ func newContainerCb(pod *pod, data []byte) error {
 		return fmt.Errorf("Container %s already exists, impossible to create", payload.ID)
 	}
 
-	absoluteRootFs, err := mountContainerRootFs(payload.ID, payload.Image, payload.RootFs, payload.FsType)
+	absoluteRootFs, err := mountContainerRootFs(payload.ID, payload.Image, payload.RootFs, payload.FsType, payload.Addr)
```
sboeuf: This function requires exclusively some `payload.*` objects, we should change the function prototype and provide `payload` only.

syscall.go
```diff
@@ -178,17 +179,76 @@ func waitForBlockDevice(deviceName string) error {
 	return nil
 }
 
-func mountContainerRootFs(containerID, image, rootFs, fsType string) (string, error) {
+func recanSCSIBus() error {
```
sboeuf: `s/recanSCSIBus/scanSCSIBus/`

syscall.go
```diff
@@ -178,17 +179,76 @@ func waitForBlockDevice(deviceName string) error {
...
+	scsiHostPath := "/sys/class/scsi_host/"
```
sboeuf: No need for last `/` for the path.

syscall.go
```diff
@@ -178,17 +179,76 @@ func waitForBlockDevice(deviceName string) error {
...
+	}
+
+	files, err := ioutil.ReadDir(scsiHostPath)
+	if err != nil {
+		return err
+	}
+
+	for _, file := range files {
+		host := file.Name()
+		scanPath := filepath.Join(scsiHostPath, host, "/scan")
```
sboeuf: No need for `"/scan"`, a regular `"scan"` is enough and you should put that into a `const` (even if it is used only here).

syscall.go
```diff
@@ -178,17 +179,76 @@ func waitForBlockDevice(deviceName string) error {
...
+		data := []byte("- - -")
```
sboeuf: Please put that into a `const` with a comprehensive name. Also add few comments to detail the meaning of this `"- - -"`

syscall.go
```diff
@@ -178,17 +179,76 @@ func waitForBlockDevice(deviceName string) error {
...
+	}
+
+	return nil
+}
+
+// findSCSIDisk finds the SCSI disk name associated with the given SCSI address
+// This approach eliminates the need to predict the disk name on the host side,
+// but we do need to rescan SCSI bus for this.
+func findSCSIDisk(scsiAddr string) (string, error) {
+	scsiPath := fmt.Sprintf("/sys/class/scsi_disk/0:0:%s/device/block/", scsiAddr)
```
sboeuf: No need for last `/`.

api/commands.go
```diff
@@ -156,6 +156,7 @@ type NewContainer struct {
 	FsType  string  `json:"fstype,omitempty"`
 	Fsmap   []Fsmap `json:"fsmap"`
 	Process Process `json:"process"`
+	Addr    string  `json:"addr,omitempty"`
```
sboeuf: Ok here I think we should go with something more comprehensive like `expectedName`, meaning this is the identifier/name that we expect to find inside the VM to find the appropriate device.

syscall.go
```diff
@@ -178,17 +179,76 @@ func waitForBlockDevice(deviceName string) error {
...
 		return "", err
 	}
 
 	var source string
 	if fsType != "" {
-		source = filepath.Join(devPath, image)
-		if err := waitForBlockDevice(image); err != nil {
-			return "", err
+		// If SCSI adddress is provided, use that to find SCSI disk
+		if scsiAddr != "" {
```
sboeuf, line 248: You should rely on `fsType` to get something like `"scsi"`. In case you determine this is a SCSI device, then you check the `expectedName`, rescan the bus and find the device, and after that you can reassign `fsType` to `bind`.

sboeuf (changes requested):

agent.go
amshinde: @sboeuf Not passing the payload here as I did not want to introduce any hypertstart data structures in syscall.go

api/commands.go
amshinde: I have changed this to SCSI address. SCSI address is a string in the format SCSI-Id:LUN. This scsi-id and lun are selected and passed from the host side. I have eliminated the expected name completely with this, to avoid any errors in computing the expected name.

syscall.go
amshinde: fixed

syscall.go
amshinde: fixed

syscall.go
amshinde: fixed

syscall.go
amshinde: I do not understand, fstype is supposed to container the file system type of the block device.
In case fstype is non-empty and scsi-addr is non-empty, we try to find the device name based on the scsi address sent. In case scsi-addr is empty but fstype is populated, we fall back to virtio-block and rely on the expected name passed.

syscall.go
amshinde: fixed

syscall.go
amshinde: i have added in my comments

PR #188, opened 2018-01-05
capabilities: Add support for Linux capabilities

amshinde:
We were using a default set of capabilities for a container
process. Instead introduce a new "Capabilities" structure
for a "Process", that will accept the capabilities
that will be applied to a process.

Signed-off-by: Archana Shinde <archana.m.shinde@intel.com>

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -84,47 +84,6 @@ const (
 	killProcessTimeout = defaultTimeout
 )
 
-var capsList = []string{
-	"CAP_AUDIT_CONTROL",
-	"CAP_AUDIT_READ",
-	"CAP_AUDIT_WRITE",
-	"CAP_BLOCK_SUSPEND",
-	"CAP_CHOWN",
-	"CAP_DAC_OVERRIDE",
-	"CAP_DAC_READ_SEARCH",
-	"CAP_FOWNER",
-	"CAP_FSETID",
-	"CAP_IPC_LOCK",
-	"CAP_IPC_OWNER",
-	"CAP_KILL",
-	"CAP_LEASE",
-	"CAP_LINUX_IMMUTABLE",
-	"CAP_MAC_ADMIN",
-	"CAP_MAC_OVERRIDE",
-	"CAP_MKNOD",
-	"CAP_NET_ADMIN",
-	"CAP_NET_BIND_SERVICE",
-	"CAP_NET_BROADCAST",
-	"CAP_NET_RAW",
-	"CAP_SETGID",
-	"CAP_SETFCAP",
-	"CAP_SETPCAP",
-	"CAP_SETUID",
-	"CAP_SYS_ADMIN",
-	"CAP_SYS_BOOT",
-	"CAP_SYS_CHROOT",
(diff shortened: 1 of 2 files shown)


agent.go
```diff
@@ -84,64 +84,6 @@ const (
...
-	"CAP_SYS_PTRACE",
-	"CAP_SYS_RAWIO",
-	"CAP_SYS_RESOURCE",
-	"CAP_SYS_TIME",
-	"CAP_SYS_TTY_CONFIG",
-	"CAP_SYSLOG",
-	"CAP_WAKE_ALARM",
-}
-
-var defaultCapsList = []string{
```
sboeuf: @amshinde don't you think we should keep the list of default in case we don't have any caps provided by the runtime. I am saying that because if we don't have a minimum set of caps, libcontainer will not be able to run our containers properly.

sboeuf (commented):
Change looks good, but I have one comment about the default caps. We should be able to fall back to a default set if nothing is provided.

agent.go
jodh-intel: We could also consider using `capability.List()` converted to a string slice as that is already vendored?

amshinde: @sboeuf I was considering that myself if we should consider falling back to a default case, I avoided that as it would mask any bugs/incompatibities in our capabilities handling from the spec.
I suppose we could log that, and move forward with the default list.

sboeuf: @amshinde no you're right we should not expect the agent to apply some caps if nothing is provided, we only want to apply the caps from the runtime.
Until now, we didn't have any mechanism to pass the caps from the runtime to the agent, but you've introduced it with https://github.com/containers/virtcontainers/pull/542 and we should exclusively rely on that now.
The user will be responsible for providing the appropriate caps according to the workload he wants to run.
One point though, if we discover there is some caps that we need because we are in the particular case of a VM in Clear Containers, those caps would be the default one that we would really need to apply.

sboeuf:
LGTM

[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/188/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent) [Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/188/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

sboeuf (approved):

amshinde:
@sboeuf @jcvenegas The jenkins 16.04 CI is failing, looks like the below issue:
https://github.com/clearcontainers/runtime/issues/896

The stdout/stderr is never sent on the tty channel in case of exec. I have seen the CI failing quite a bit because of this. I am not sure what has changed in the exec code flow.

On the other hand, I would like to get this change merged since the limited capability set breaks quite a few things correctly.

sboeuf:
@amshinde I agree we should merge your change since the instability is not introduced by your change. I have restarted the CI for 16.04 because I think this should pass even if the passing rate is not 100% successful.
In any case, we'll investigate this issue if it shows up more often.

PR #190, opened 2018-01-15
agent: Add support to online vCPUs

### LINKED ISSUE #189: Support hot plugged vCPUs
Online vCPUs when they are hot plugged

devimc:
this patch is to online vCPU in the POD when a new vCPU is
hot added by the runtime

fixes #189

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -72,6 +72,7 @@ const (
 	cannotGetTimeMsg   = "Failed to get time for event %s:%v"
 	pciBusRescanFile   = "/sys/bus/pci/rescan"
 	pciBusMode         = 0220
+	cpuMode            = 0600
 
 	// If a process terminates because of signal "n"
 	// The exit code is "128 + signal_number"
@@ -82,6 +83,10 @@ const (
 	defaultTimeout     = 15
 	runProcessTimeout  = defaultTimeout
 	killProcessTimeout = defaultTimeout
+
+	// Udev subsystem
+	blockSubsystem = "block"
+	cpuSubsystem   = "cpu"
 )
 
 var defaultCapsList = []string{
@@ -531,7 +536,12 @@ func (p *pod) listenToUdevEvents(done chan struct{}) {
 
 			// Ignore udev events for block devices, these are handled by another go-routine in waitForBlockDevice
 			// that mounts block devices at the expected location provided on the command line.
-			if d.Subsystem() == "block" {
+			if d.Subsystem() == blockSubsystem {
+				continue
+			}
+
+			if d.Subsystem() == cpuSubsystem {
+				p.onlineVCPU(d.Syspath())
 				continue
 			}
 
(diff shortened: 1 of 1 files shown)


agent.go
```diff
@@ -72,6 +72,7 @@ const (
 	cannotGetTimeMsg   = "Failed to get time for event %s:%v"
 	pciBusRescanFile   = "/sys/bus/pci/rescan"
 	pciBusMode         = 0220
+	cpuMode            = 0644
```
sboeuf: I think `0600` should be enough.

agent.go
```diff
@@ -576,6 +577,11 @@ func (p *pod) listenToUdevEvents(done chan struct{}) {
 				continue
 			}
 
+			if d.Subsystem() == "cpu" {
```
sboeuf: Might worth putting `"block"` `"cpu"` and `"add"` into some constants.

agent.go
jodh-intel: Yeah, maybe even `0200`?

agent.go
```diff
@@ -593,6 +599,16 @@ func (p *pod) listenToUdevEvents(done chan struct{}) {
 	}()
 }
 
+func (p *pod) onlineVCPU(syspath string) {
+	onlinePath := filepath.Join(syspath, "online")
```
sboeuf, line 566: I assume the syspath you're gonna get here is going to be something like:
```
/sys/devices/system/cpu/cpuN
```
Right ?
Could be nice to check the path is the expected pattern before to concatenate it with `"online"`.

agent.go
devimc: this file is used to set and get the cpu status, ```0600``` should be enough

agent.go
```diff
@@ -593,6 +599,16 @@ func (p *pod) listenToUdevEvents(done chan struct{}) {
...
+	if err := ioutil.WriteFile(onlinePath, []byte("1"), cpuMode); err != nil {
+		fieldLogger := agentLog.WithField("subsystem", "udevlistener")
+		fieldLogger.WithFields(logrus.Fields{
```
sboeuf: You can simplify this with something like that:
```go
agentLog.WithFields(logrus.Fields{
        "subsystem": "udevlistener",
        "subsystem": "cpu",
}).Errorf("failed to online vCPU %s: %s", syspath, err)
```

sboeuf (commented):
A few comments but the change looks good.

agent.go
devimc: I'd prefer to just check if the file exist

devimc:
@sboeuf  @jodh-intel  changes applied, thanks

agent.go
sboeuf: Sounds good.

sboeuf (approved):

sboeuf:
LGTM

jodh-intel:
lgtm

[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/190/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

jodh-intel:
Restarted 17.04 CI - looks like a network timeout talking to the Ubuntu servers.

agent.go
```diff
@@ -531,7 +536,12 @@ func (p *pod) listenToUdevEvents(done chan struct{}) {
 
 			// Ignore udev events for block devices, these are handled by another go-routine in waitForBlockDevice
 			// that mounts block devices at the expected location provided on the command line.
-			if d.Subsystem() == "block" {
+			if d.Subsystem() == blockSubsystem {
+				continue
+			}
+
+			if d.Subsystem() == cpuSubsystem {
+				p.onlineVCPU(d.Syspath())
```
amshinde, line 544: if this operation takes some time, you can consider doing it in a separate go-routine.

devimc: this function only writes ```1``` in ```/sys/devices/system/cpu/cpuX/online```

PR #197, opened 2018-01-20
agent: Fix sync issues with exit code and stdout of a process

### LINKED ISSUE #196: Agent sometimes return exit code before stdout
The agent mixes the order between the exit code and the stdout of a process sometimes. When the exit code is received before the stdout by the shim, the stdout is dropped because the shim does not expect something else to come.

sboeuf:
The patch about subreaping introduced some bugs by removing some
code that was actually very important. This code relates to the
synchronization of the stdout coming from the process and its exit
code being returned before. In case the exit code was returned before
stdout was not forwarded to the shim since the connection had been
dropped.

This commit restores few lines of code ensuring all output has been
forwarded before to send the exit code.

Fixes #196

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -1031,6 +1031,12 @@ func (p *pod) runContainerProcess(cid, pid string, terminal bool, started chan e
 		return err
 	}
 
+	// Make sure all output has been routed before to go further and
+	// return the exit code.
+	ctr.closeProcessStreams(pid)
+	ctr.closeProcessPipes(pid)
+	wgRouteOutput.Wait()
+
 	agentLog.WithField("exit-code", exitCode).Info("got wait exit code")
 
 	// Send empty message on tty channel to close the IO stream


agent.go
```diff
@@ -1031,6 +1031,12 @@ func (p *pod) runContainerProcess(cid, pid string, terminal bool, started chan e
 		return err
 	}
 
+	// Make sure all output has been routed before to go further and
+	// return the exit code.
+	ctr.closeProcessStreams(pid)
```
amshinde, line 1036: @sboeuf do we need the closeProcessStreams? Its been called in defer.

sboeuf: If we don't, `wgRouteOutput.Wait()` would be stuck forever. The design is not the prettiest for this function but it works.
And the defer is not a problem cause if the streams have been closed and set to nil, they won't be closed again.
We could (actually should) improve this code.

jcvenegas:
I will test this today and check the issues I've seen are fixed now. Thanks for the fix :)

sboeuf:
@jcvenegas any news about this ? Have you been able to test it ?
@amshinde should not we merge this at least to get back to the state where we were before ?

amshinde:
@sboeuf Yeah , I agree we should merge this.

amshinde (approved):
lgtm

PR #205, opened 2018-01-25
Add support for scsi

amshinde:
Add support for SCSI disks to be passed for container rootfs and volumes.
If SCSI address is provided, use that to get the SCSI disk name.
This eliminates the need to predict the drive name of the block
device on the host side. We do need to incur the cost of scanning SCSI bus.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -1306,6 +1306,15 @@ func addMounts(config *configs.Config, fsmaps []hyper.Fsmap) error {
 	for _, fsmap := range fsmaps {
 
 		source := fsmap.Source
+		var err error
+
+		if fsmap.SCSIAddr != "" {
+			source, err = getSCSIDisk(fsmap.SCSIAddr)
+			if err != nil {
+				return err
+			}
+		}
+
 		if !fsmap.AbsolutePath {
 			source = filepath.Join(mountShareDirDest, fsmap.Source)
 		}
--- a/agent.go
+++ b/agent.go
@@ -1387,7 +1387,7 @@ func newContainerCb(pod *pod, data []byte) error {
 		return fmt.Errorf("Container %s already exists, impossible to create", payload.ID)
 	}
 
-	absoluteRootFs, err := mountContainerRootFs(payload.ID, payload.Image, payload.RootFs, payload.FsType)
+	absoluteRootFs, err := mountContainerRootFs(payload)
 	if err != nil {
 		return err
 	}
--- a/api/commands.go
+++ b/api/commands.go
@@ -192,6 +192,9 @@ type NewContainer struct {
 	Process          Process          `json:"process"`
 	SystemMountsInfo SystemMountsInfo `json:"systemMountsInfo"`
 	Constraints      Constraints      `json:"constraints"`
+
+	// SCSI address in the format SCSI-Id:LUN
+	SCSIAddr string `json:"scsiAddr,omitempty"`
 }
 
 // KillContainer describes the format expected by a KILLCONTAINER command.
--- a/api/commands.go
+++ b/api/commands.go
@@ -110,6 +110,11 @@ type Fsmap struct {
 	AbsolutePath bool   `json:"absolutePath"`
 	ReadOnly     bool   `json:"readOnly"`
 	DockerVolume bool   `json:"dockerVolume"`
+
+	// SCSIAddr is the SCSI address in the format SCSI-id:LUN.
+	// If SCSIAddr is provided, we use that to determine the device name to be mounted
+	// ignoring the Source field that is used for as the source location for mounting otherwise.
+	SCSIAddr string `json:"scsiAddr,omitempty"`
 }
 
 // Capabilities specify the capabilities to keep when executing the process inside the container.
--- a/syscall.go
+++ b/syscall.go
@@ -18,11 +18,14 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 	"syscall"
 	"time"
 
+	hyper "github.com/clearcontainers/agent/api"
 	goudev "github.com/jochenvg/go-udev"
 	"github.com/sirupsen/logrus"
 )
@@ -115,10 +118,25 @@ func unmountShareDir() error {
 	return os.RemoveAll(containerMountDest)
 }
 
-func waitForBlockDevice(deviceName string) error {
-	devicePath := filepath.Join(devPath, deviceName)
+var (
+	// Here in "0:0", the first number is the SCSI host number, while the second number is always 0.
+	scsiHostChannel = "/0:0:"
 
-	if _, err := os.Stat(devicePath); err == nil {
+	scsiDiskPrefix = "/sys/class/scsi_disk" + scsiHostChannel
+
+	scsiDiskSuffix = "/device/block"
+)
+
+func waitForBlockDevice(deviceName string, isSCSIAddr bool) error {
+	var path string
+
+	if isSCSIAddr {
+		path = filepath.Join(scsiDiskPrefix+deviceName, scsiDiskSuffix)
+	} else {
+		path = filepath.Join(devPath, deviceName)
+	}
+
+	if _, err := os.Stat(path); err == nil {
 		return nil
 	}
 
@@ -147,19 +165,24 @@ func waitForBlockDevice(deviceName string) error {
 		fieldLogger.Info("Started listening for udev events for block device hotplug")
 
 		// Check if the device already exists.
-		if _, err := os.Stat(devicePath); err == nil {
+		if _, err := os.Stat(path); err == nil {
 			fieldLogger.Info("Device already hotplugged, quit listening")
 		} else {
 
 			for d := range ch {
 				fieldLogger = fieldLogger.WithFields(logrus.Fields{
 					"udev-path":  d.Syspath(),
+					"dev-path":   d.Devpath(),
 					"udev-event": d.Action(),
 				})
 
 				fieldLogger.Info("got udev event")
-				if d.Action() == "add" && filepath.Base(d.Devpath()) == deviceName {
-					fieldLogger.Info("Hotplug event received")
+
+				if isSCSIAddr && d.Action() == "add" && strings.Contains(d.Devpath(), scsiHostChannel+deviceName+"/block") {
+					fieldLogger.Info("SCSI hotplug event received")
+					break
+				} else if d.Action() == "add" && filepath.Base(d.Devpath()) == deviceName {
+					fieldLogger.Info("Block hotplug event received")
 					break
 				}
 			}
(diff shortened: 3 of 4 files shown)


syscall.go
```diff
@@ -178,21 +193,115 @@ func waitForBlockDevice(deviceName string) error {
...
+		return fmt.Errorf("Unexpected format for SCSI Address : %s, expect SCSIID:LUN", scsiAddr)
+	}
+
+	// Scan scsi host passing in the channel, SCSI id and LUN, channel is always 0
+	scanData := []byte(fmt.Sprintf("0 %s %s", tokens[0], tokens[1]))
+
+	for _, file := range files {
+		host := file.Name()
+		scanPath := filepath.Join(scsiHostPath, host, "scan")
+		if err := ioutil.WriteFile(scanPath, scanData, 0666); err != nil {
```
jodh-intel: This should be `0200` I think.

syscall.go
```diff
@@ -178,21 +193,115 @@ func waitForBlockDevice(deviceName string) error {
...
+
+	return nil
+}
+
+// findSCSIDisk finds the SCSI disk name associated with the given SCSI address
+// This approach eliminates the need to predict the disk name on the host side,
+// but we do need to rescan SCSI bus for this.
+func findSCSIDisk(scsiAddr string) (string, error) {
+	// Here in "0:0", the first number is the SCSI host number, while the second number is always 0.
+	scsiPath := fmt.Sprintf("/sys/class/scsi_disk/0:0:%s/device/block", scsiAddr)
```
jodh-intel: This path prefix is also used in `waitForBlockDevice()`. Can you make it a file-global `var` as that will avoid the duplication and hard-coding, but also allow the value to be changed for testing.

agent.go
```diff
@@ -1214,10 +1214,19 @@ func addMounts(config *configs.Config, fsmaps []hyper.Fsmap) error {
...
 
 		source := fsmap.Source
+		var err error
+
 		if !fsmap.AbsolutePath {
 			source = filepath.Join(mountShareDirDest, fsmap.Source)
 		}
 
+		if fsmap.SCSIAddr != "" {
+			source, err = getSCSIDisk(fsmap.SCSIAddr)
```
jodh-intel: - This is resetting `source` so is it worth adding an extra check to see if `source != ""` here? Or is it guaranteed that if `SCSIAddr` is set that `Source` is irrelevant?

  Could you maybe add some comments to the `Fsmap` type to explain this?

- Does this check need to come before the `AbsolutePath` check above as currently that code is being executed for no reason if we're using a SCSI device?

syscall.go
```diff
@@ -115,10 +117,18 @@ func unmountShareDir() error {
...
 }
 
-func waitForBlockDevice(deviceName string) error {
+func waitForBlockDevice(deviceName string, isSCSIAddr bool) error {
 	devicePath := filepath.Join(devPath, deviceName)
 
-	if _, err := os.Stat(devicePath); err == nil {
+	if isSCSIAddr {
+		// Here in "0:0", the first number is the SCSI host number, while the second number is always 0.
+		scsiPrefix := "/sys/class/scsi_disk/0:0"
```
jodh-intel: Can you make this path prefix a file-global `var` to allow it to be re-used (and changed for testing)?

syscall.go
```diff
@@ -115,10 +117,18 @@ func unmountShareDir() error {
...
+
+		path := scsiPrefix + deviceName + "/device/block"
```
jodh-intel: Same comment as above.

syscall.go
```diff
@@ -115,10 +117,18 @@ func unmountShareDir() error {
...
+		if _, err := os.Stat(path); err == nil {
+			return nil
+		}
+	} else if _, err := os.Stat(devicePath); err == nil {
```
jodh-intel: I think it would be clearer if you moved the `devicePath` variable down to here and removed the `else if`:

```
if isSCSIAddr {
    :
}

devicePath := filepath.Join(devPath, deviceName)

if _, err := os.Stat(devicePath); err == nil {
    return nil
}


```

syscall.go
```diff
@@ -154,11 +164,16 @@ func waitForBlockDevice(deviceName string) error {
...
 					"udev-path":  d.Syspath(),
+					"dev-path":   d.Devpath(),
 					"udev-event": d.Action(),
 				})
 
 				fieldLogger.Info("got udev event")
-				if d.Action() == "add" && filepath.Base(d.Devpath()) == deviceName {
+
+				if isSCSIAddr && d.Action() == "add" && strings.Contains(d.Devpath(), deviceName) {
+					fieldLogger.Info("Hotplug event received")
```
jodh-intel: Maybe "SCSI hotplug event received"?

syscall.go
```diff
@@ -178,21 +193,115 @@ func waitForBlockDevice(deviceName string) error {
...
+		scanPath := filepath.Join(scsiHostPath, host, "scan")
+		if err := ioutil.WriteFile(scanPath, scanData, 0666); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// findSCSIDisk finds the SCSI disk name associated with the given SCSI address
```
jodh-intel: Nit: Missing period at end of sentence.

syscall.go
```diff
@@ -178,21 +193,115 @@ func waitForBlockDevice(deviceName string) error {
...
+	if _, err := os.Stat(scsiPath); err != nil {
+		return "", err
+	}
+
+	files, err := ioutil.ReadDir(scsiPath)
+	if err != nil {
+		return "", err
+	}
+
+	if len(files) > 1 {
```
jodh-intel: There is an implicit assumption that a single file was read, so how about the following:

```
l := len(files)

if l == 0 || l > 1 {
    ...
}
```

jodh-intel (changes requested):
Hi @amshinde - thanks for raising. I've added a few comments.

Please could you add some tests for these changes as well?

agent.go
```diff
@@ -1295,7 +1295,7 @@ func newContainerCb(pod *pod, data []byte) error {
 		return fmt.Errorf("Container %s already exists, impossible to create", payload.ID)
 	}
 
-	absoluteRootFs, err := mountContainerRootFs(payload.ID, payload.Image, payload.RootFs, payload.FsType)
+	absoluteRootFs, err := mountContainerRootFs(payload.ID, payload.Image, payload.RootFs, payload.FsType, payload.SCSIAddr)
```
sboeuf: We should pass the payload to this function, instead of extending its number of parameters, all being attributes of the same structure.

api/commands.go
```diff
@@ -182,6 +182,9 @@ type NewContainer struct {
 	Fsmap            []Fsmap          `json:"fsmap"`
 	Process          Process          `json:"process"`
 	SystemMountsInfo SystemMountsInfo `json:"systemMountsInfo"`
+
+	// SCSI address in the format SCSI-Id:LUN
+	SCSIAddr string `json:"scsiAddr,omitempty"`
```
sboeuf: I am not sure this is the right place for this field. This is very specific to a type of rootfs. Maybe we should define a structure Rootfs so that we can actually include this type of information. WDYT ?

syscall.go
```diff
@@ -115,10 +117,18 @@ func unmountShareDir() error {
 	return os.RemoveAll(containerMountDest)
 }
 
-func waitForBlockDevice(deviceName string) error {
+func waitForBlockDevice(deviceName string, isSCSIAddr bool) error {
```
sboeuf: You should rename this function into `waitForDevice()` and pass a more global structure. Once again, this looks very specific to SCSI here.

sboeuf (changes requested):

api/commands.go
amshinde: Yes, this is specific to rootfs, but we have other fields here like Rootfs, FsMap and FsType that are related to rootfs that Iwould be candidates for a combined structure. I would like to do that in a separate PR, as this involves a lot of components.

agent.go
amshinde: I had done this to avoid introducing hyperstart specific structures in syscall.go.

syscall.go
amshinde: Done

syscall.go
amshinde: done

syscall.go
amshinde: fixed

agent.go
amshinde: I have added in comments.

syscall.go
amshinde: Done. Have added tests as well.

syscall.go
amshinde: Yes, it should be 0200. Fixed.

syscall.go
```diff
@@ -115,10 +117,19 @@ func unmountShareDir() error {
...
-	devicePath := filepath.Join(devPath, deviceName)
+// Here in "0:0", the first number is the SCSI host number, while the second number is always 0.
+var scsiDiskPrefix = "/sys/class/scsi_disk/0:0:"
 
-	if _, err := os.Stat(devicePath); err == nil {
+func waitForBlockDevice(deviceName string, isSCSIAddr bool) error {
+	var path string
+
+	if isSCSIAddr {
+		path = scsiDiskPrefix + deviceName + "/device/block"
```
jodh-intel: This would be more consistent (and maybe safer) as:

```
path = filepath.Join(scsiDiskPrefix + deviceName, "/device/block")
```

You could also create a variable for the `/device/block` which is used multiple times.

syscall.go
```diff
@@ -147,19 +158,25 @@ func waitForBlockDevice(deviceName string) error {
...
 					"udev-path":  d.Syspath(),
+					"dev-path":   d.Devpath(),
 					"udev-event": d.Action(),
 				})
 
 				fieldLogger.Info("got udev event")
-				if d.Action() == "add" && filepath.Base(d.Devpath()) == deviceName {
-					fieldLogger.Info("Hotplug event received")
+
+				//if isSCSIAddr && d.Action() == "add" && strings.Contains(d.Devpath(), deviceName) {
```
jodh-intel: Commented out line.

syscall.go
```diff
@@ -147,19 +158,25 @@ func waitForBlockDevice(deviceName string) error {
...
+				if isSCSIAddr && d.Action() == "add" && strings.Contains(d.Devpath(), "/0:0:"+deviceName+"/block") {
```
jodh-intel: Can you create a global private variable for the `0:0` (and ideally change `scsiDiskPrefix` to use it too)?

syscall.go
```diff
@@ -178,21 +195,119 @@ func waitForBlockDevice(deviceName string) error {
 	return nil
 }
 
-func mountContainerRootFs(containerID, image, rootFs, fsType string) (string, error) {
+var scsiHostPath = "/sys/class/scsi_host"
+
+// scanSCSIBus scans SCSI bus for the given SCSI address(SCSI-Id and LUN)
+func scanSCSIBus(scsiAddr string) error {
+	if _, err := os.Stat(scsiHostPath); err != nil {
```
jodh-intel: This call isn't needed as the `ReadDir()` below will detect `ENOENT` itself. Same comment for the `Stat()` in `findSCSIDisk()`.

syscall.go
```diff
@@ -178,21 +195,119 @@ func waitForBlockDevice(deviceName string) error {
...
+
+	if err == nil {
+		if err = waitForBlockDevice(scsiDiskName, false); err != nil {
+			return "", err
+		}
+	} else {
+		// If device node is not found, wait for udev event for the scsi disk,
+		// with the format "0:SCSIId:LUN/block"
+
+		//if err := waitForBlockDevice("0:"+scsiAddr+"/block", true); err != nil {
```
jodh-intel: Can you remove this line please.

amshinde:
@jodh-intel I have pushed changes for your comments. Please have a look.

agent.go
sboeuf: There is no harm doing so. `syscall.go` is part of the agent, which makes it tied to hyperstart. I don't think this is an issue to pass an hyperstart structure to this function, and this would simplify things as I mentioned.

api/commands.go
sboeuf: Fair enough, but please open an issue to track this down ;)

syscall.go
sboeuf: Any comment on this ?

amshinde: I would prefer keeping this as  is as we are waiting for a Block device here, the difference being if the device is passed with virtio-scsi or virtio-block, with the flag indicating this.

api/commands.go
amshinde: Sure will do.

agent.go
amshinde: I was just trying to avoid introducing any hyperstart structures here, as this is more like an utils package and is a good contender for moving into its own util package, but that is for the future. 
I can simplify this for now.

amshinde: I have pushed changes for this.

api/commands.go
amshinde: https://github.com/clearcontainers/agent/issues/210

sboeuf: thx

syscall.go
sboeuf: Yes I agree this is still a block device, forget this comment.

sboeuf (approved):

sboeuf:
LGTM
@jodh-intel @egernst please make a new review !

jodh-intel (approved):

PR #208, opened 2018-02-09
process: Fix the way output is handled after process termination

### LINKED ISSUE #207: process: Make sure we don't miss any output
The way the agent code was written, it was forcing the end of the output (stdout/stderr) right after the process was returning. Instead, by properly closing all the remaining pipes, we should be able to wait for the proper end of stdout/stderr. This way, we won't miss any output before to return the exit code to the shim.

Refers to https://github.com/clearcontainers/runtime/issues/997

sboeuf:
This PR does a bit of cleanup, but it mainly fixes the way the agent should properly wait for the end of any process output, instead of forcing the closure of its output.

Fixes #207

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -989,10 +989,10 @@ func (p *pod) runContainerProcess(cid, pid string, terminal bool, started chan e
 	ctr := p.getContainer(cid)
 
 	defer func() {
-		ctr.wgProcesses.Done()
-		ctr.deleteProcess(pid)
-		ctr.closeProcessStreams(pid)
 		ctr.closeProcessPipes(pid)
+		ctr.closeProcessStreams(pid)
+		ctr.deleteProcess(pid)
+		ctr.wgProcesses.Done()
 	}()
 
 	var wgRouteOutput sync.WaitGroup
--- a/agent.go
+++ b/agent.go
@@ -952,6 +952,35 @@ func (c *container) closeProcessPipes(pid string) {
 	}
 }
 
+func (c *container) closeConsoleSocket(pid string) error {
+	cid := c.container.ID()
+	proc := c.getProcess(pid)
+
+	errMsg := "Could not close process socket pair"
+
+	if proc == nil {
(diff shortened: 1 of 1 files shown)


sboeuf:
Fixes https://github.com/clearcontainers/runtime/issues/997

grahamwhaley:
@sboeuf do you think this is also related to https://github.com/kata-containers/agent/issues/125 ?

agent.go
```diff
@@ -952,6 +952,35 @@ func (c *container) closeProcessPipes(pid string) {
...
+
+	errMsg := "Could not close process socket pair"
+
+	if proc == nil {
+		return fmt.Errorf("Container %s process %s no longer exists", cid, pid)
+	}
+
+	if proc.consoleSock != nil {
+		if err := proc.consoleSock.Close(); err != nil {
+			return fmt.Errorf(errMsg)
```
jodh-intel: You don't need `Errorf()` as written. But how about making it as follows (here and below)?

```
return fmt.Errorf("%s: %s", errMsg, err)
```

jodh-intel: (By that I mean you could just use `errors.New(errMsg)`).

sboeuf: Yes let me change that ;)

sboeuf:
@grahamwhaley no I don't think this is related. The whole logic behind getting stdout/stderr is handled differently with kata agent.

sboeuf:
@jodh-intel PR updated.

jodh-intel (approved):
lgtm

grahamwhaley (approved):
Nicely broken out and written patches btw.
lgtm

sboeuf:
@grahamwhaley thx ;)

jcvenegas:
lgtm
agree with @grahamwhaley, makes easy to get the root cause. 


[Approved with PullApprove](https://pullapprove.com/clearcontainers/agent/pull-request/208/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=clearcontainers/agent)

PR #215, opened 2018-03-08
mount: Use 9p2000.L version when mounting 9p filesystem

### LINKED ISSUE #214: mounts: add `version=9p2000.L` to agent mounts
Over on #211 we showed that moving to using `9p2000.L` as the mount type fixes a couple of symlink test fails, and has no particularly visible change in performance. Make the change at https://github.com/clearcontainers/agent/blob/master/syscall.go#L110

sboeuf:
Using 9p2000.L version helps passing more tests compared to the
standard version. The symlinks are handled in a better way.

Fixes #214

### DIFF
--- a/syscall.go
+++ b/syscall.go
@@ -32,6 +32,7 @@ import (
 
 const mountPerm = os.FileMode(0755)
 const devPath = "/dev"
+const mntOptions9p = "trans=virtio,version=9p2000.L"
 
 // bindMount bind mounts a source in to a destination, with the recursive
 // flag if needed.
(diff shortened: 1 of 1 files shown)


grahamwhaley (approved):
lgtm
nice pull out to the const.
Be interesting to see what the CIs think.

syscall.go
```diff
@@ -32,6 +32,7 @@ import (
 
 const mountPerm = os.FileMode(0755)
 const devPath = "/dev"
+const 9pMntOptions = "trans=virtio,version=9p2000.L"
```
jodh-intel: This isn't valid as it starts with a numeric.

sboeuf:
@devimc @amshinde @jcvenegas PTAL

amshinde (approved):
lgtm

